PR #18, opened 2025-04-09
Reset Batcher alarm for immediate execution if there is an expired alarm

### LINKED ISSUE #11: Client timeouts and internal server errors
The write success rate to the CT logs degrades over time when the logs are under steady load. I suspect it's an issue in the [Batchers](https://github.com/cloudflare/azul/blob/main/crates/ct_worker/src/batcher_do.rs), which function pretty similarly to the [example in the alarms docs](https://developers.cloudflare.com/durable-objects/examples/alarms-api/).

In order to add artificial load to the cftest2025{h1a,h2b} logs, there's a service cross-pollinating entries from the sycamore2025{h1b,h2b} logs, respectively. After a fresh deploy which restarts the Batcher DOs, most responses are "ok" or "not accepted" (non-errors) for a short time:

`cftest2025h1a`:
```
2025/04/11 00:57:06 main.go:110: source=sycamore2025h1b target=cftest2025h1a offset=1907245404 max=1907245404 count=53 "ok"=53
```
`cftest2025h2a`:
```
2025/04/11 00:57:06 main.go:110: source=sycamore2025h2b target=cftest2025h2a offset=137315106 max=137315106 count=1607 "ok"=1607
```

A few hours later, things have degraded. Clients receive timeouts and internal server errors.

`cftest2025h1a`, only ~21% of writes succeeded:
```
2025/04/11 12:34:07 main.go:110: source=sycamore2025h1b target=cftest2025h1a offset=1907252773 max=1907252773 count=7419 "Post \"https://static-ct.cloudflareresearch.com/logs/cftest2025h1a/ct/v1/add-chain\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"=4847 "Post \"https://static-ct.cloudflareresearch.com/logs/cftest2025h1a/ct/v1/add-pre-chain\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"=8 "not accepted"=13 "ok"=1578 "unexpected response status: 429 Too Many Requests"=219 "unexpected response status: 500 Internal Server Error"=751 "unexpected response status: 502 Bad Gateway"=3
```

`cftest2025h2a`, about 80% of writes succeeded.
```
2025/04/11 12:34:06 main.go:110: source=sycamore2025h2b target=cftest2025h2a offset=137534484 max=137534484 count=22111 "Post \"https://static-ct.cloudflareresearch.com/logs/cftest2025h2a/ct/v1/add-chain\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"=1880 "not accepted"=14 "ok"=17646 "unexpected response status: 429 Too Many Requests"=2239 "unexpected response status: 500 Internal Server Error"=332
```

When in the degraded state, I see some of these errors in the logs (`wrangler tail`), which looks to be the result of exceeding the JavaScript array size supported by the platform: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length.
```
failed to submit batch: Error: RangeError: Invalid array buffer length - Cause: RangeError: Invalid array buffer length
```
And some of these:
```
POST https://static-ct.cloudflareresearch.com/logs/cftest2025h2a/ct/v1/add-chain - Ok @ 4/10/2025, 8:27:04 PM
  (warn) Internal error: Error: Error: internal error; reference = a7obba7llanokdmd7miqnr4d - Cause: Error: internal error; reference = a7obba7llanokdmd7miqnr4d
POST https://static-ct.cloudflareresearch.com/logs/cftest2025h2a/ct/v1/add-chain - Ok @ 4/10/2025, 8:26:59 PM
  (warn) Internal error: Error: Error: internal error; reference = 9lumerne5f04b5euemarcdko - Cause: Error: internal error; reference = 9lumerne5f04b5euemarcdko
```
and these:
```
RuntimeError: memory access out of bounds\n    at wasm://wasm/00cce826:wasm-function[5070]:0x1e7c3a\n    at wasm://wasm/00cce826:wasm-function[2016]:0x17d33c\n    at new z (shim.js:186:15)
```

lukevalenta:
fixes #11

### DIFF
--- a/crates/ct_worker/src/batcher_do.rs
+++ b/crates/ct_worker/src/batcher_do.rs
@@ -157,6 +157,24 @@ impl DurableObject for Batcher {
         if let Err(e) = self.submit_batch(batch).await {
             log::warn!("failed to submit batch: {e}");
         }
+
+        // It's possible that a new alarm was set and failed to fire while
+        // the current alarm() was executing, since only one instance of alarm()
+        // can run at a time per Durable Object instance.
+        // In that case, schedule the alarm for immediate execution.
+        if self
+            .state
+            .storage()
+            .get_alarm()
+            .await?
+            .is_some_and(|t| t < i64::try_from(util::now_millis()).unwrap())
+        {
+            self.state
+                .storage()
+                .set_alarm(Duration::from_millis(0))
+                .await?;
+        }
+
         Response::empty()
     }
 }


crates/ct_worker/src/batcher_do.rs
```diff
@@ -157,6 +157,24 @@ impl DurableObject for Batcher {
...
+        // It's possible that a new alarm was set and failed to fire while
+        // the current alarm() was executing, since only one instance of alarm()
+        // can run at a time per Durable Object instance.
+        // In that case, schedule the alarm for immediate execution.
+        if self
+            .state
+            .storage()
+            .get_alarm()
+            .await?
+            .is_some_and(|t| t < i64::try_from(util::now_millis()).unwrap())
```
cjpatton, line 170: It looks like this says that if the time at which the alarm fired is less than the current time, then set the alarm. This sound correct, but I just wanted make sure my interpretation was correct.

Also, worth checking that `t` is miliseconds and not seconds.

Also, any reason to unwrap rather than handle the error gracefully?

cjpatton (approved):

crates/ct_worker/src/batcher_do.rs
lukevalenta: Yes, correct! Although, it doesn't look like this fix actually resolves the issues I was seeing--the test logs are still returning 500s and timeouts. And I confirmed with the Workers devs that the second alarm will already fire immediately after the first one finishes executing so this code likely has no effect. I'll decline this PR and keep hunting for the source of #11.

Reason to unwrap instead of gracefully handling the error is that `util::now_millis()` isn't expected to overflow an `i64` until some distant point in the future, so if the conversion throws an error it's a bug in the underlying platform. Based on https://burntsushi.net/unwrap/, that's a reasonable use for `unwrap`.

PR #22, opened 2025-04-15
Fix memory leak and other improvements

lukevalenta:
Remove Batcher memory cache. There's a known bug
(https://github.com/cloudflare/workers-rs/issues/722) where
wasm-bindgen's heap doesn't get cleaned up when the Durable Object is
evicted, which can happen anytime (but usually ~10s inactivity).  To fix
this, we can remove the Batcher's heap-allocated memory cache, and
instead use the batch's done channel to return sequenced entries to
waiting requests. There's still some potential for memory leaks, but
this should help quite a bit. Note that the Sequencer's memory cache has
the same problem, but should be evicted much less frequently as it's
kept alive with an alarm. Fixes https://github.com/cloudflare/azul/issues/11.

Replace Batcher alarm. As requests are kept open, we can use Delay to
provide the same batching functionality. As an added benefit, we no
longer need to use Storage at all for the Batcher.

Add 'processed' counter to track how long a Batcher Durable Object stays
alive.

See this screenshot for memory usage before/after.
Screenshot 2025-04-14 at 8 16 35 PM

### DIFF
--- a/crates/ct_worker/src/batcher_do.rs
+++ b/crates/ct_worker/src/batcher_do.rs
@@ -2,17 +2,18 @@
 // Licensed under the BSD-3-Clause license found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
 
 //! Batcher buffers entries to be sequenced, submits batches to the sequencer, and
-//! waits until the entries are sequenced. Each time the Batcher gets an update with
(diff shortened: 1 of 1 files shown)


thibmeu (approved):
Looks ok. I dd not test locally, but have seen the results

PR #23, opened 2025-04-15
Switch to SQLite backend, fixes #2

### LINKED ISSUE #2: Switch to SQLite backend for DO storage
After <https://github.com/cloudflare/workers-rs/issues/645>, switch to using SQLite API for DO storage. This might make it more feasible to write the staging bundle to DO storage instead of R2 (as long as we break up into smaller chunks to stay within [size limits](https://developers.cloudflare.com/durable-objects/platform/limits/)).

lukevalenta:
The SQLite backend supports the KV API, so there's no need to wait for https://github.com/cloudflare/workers-rs/issues/645.

We'll need to re-create the cftest logs to update the backend if to start using it now, but otherwise they'll automatically be migrated sometime in 2025: https://blog.cloudflare.com/sqlite-in-durable-objects/#how-do-i-use-it.

I confirmed that I was able to deploy this to my Free Cloudflare account.

### DIFF
--- a/crates/ct_worker/wrangler.jsonc
+++ b/crates/ct_worker/wrangler.jsonc
@@ -50,7 +50,7 @@
                 {
                     // tag should be unique for each entry
                     "tag": "v1",
-                    "new_classes": [
+                    "new_sqlite_classes": [
                         "Sequencer",
                         "Batcher"
                     ]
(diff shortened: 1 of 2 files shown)


crates/ct_worker/wrangler.jsonc
```diff
@@ -50,7 +50,7 @@
                 {
                     // tag should be unique for each entry
                     "tag": "v1",
-                    "new_classes": [
+                    "new_sqlite_classes": [
```
thibmeu, line 53: dammit, this is so new

thibmeu (approved):

PR #24, opened 2025-04-15
Reduce Wasm binary size and use start function for initialization

lukevalenta:
Simple changes to reduce code size. See individual commits for detailed descriptions.


thibmeu (approved):
i think i've seen these changes in a previous PR, and they're ok. thanks!

lukevalenta:
> i think i've seen these changes in a previous PR, and they're ok. thanks!

Yep, just moved those out of the other PR to make it easier to review. Thanks!

PR #35, opened 2025-05-13
Clean up and refactor in preparation for tlog CA

lukevalenta:
* a418754 Unify errors in static_ct_api crate
* d869cb4 Add crate for reusable x509 chain operations
* ef99499 Add crate for reading and writing big-endian length-prefixed bytes
* d06bebd Move compute_cache_hash to LogEntry::lookup_key
* 7f8176d Rename CacheKey->LookupKey and CacheValue->SequenceMetadata
* bad5891 Remove wrapper around load_pem_chain from static_ct_api
* 6643dc8 Add units for config default functions

### DIFF
--- a/crates/ct_worker/Cargo.toml
+++ b/crates/ct_worker/Cargo.toml
@@ -29,47 +29,47 @@ crate-type = ["cdylib"]
 [build-dependencies]
 chrono.workspace = true
 config = { path = "./config", package = "ct_worker_config" }
-jsonschema = "0.26.2"
+jsonschema.workspace = true
 serde_json.workspace = true
 serde.workspace = true
-url = "2.5.4"
+url.workspace = true
 x509-verify.workspace = true
 
 [dev-dependencies]
 rand = { workspace = true, features = ["small_rng"]}
-itertools = "0.13.0"
-parking_lot = "0.11"
-futures-executor = "0.3.31"
+itertools.workspace = true
+parking_lot.workspace = true
+futures-executor.workspace = true
 
 [dependencies]
 anyhow.workspace = true
 base64.workspace = true
 byteorder.workspace = true
 config = { path = "./config", package = "ct_worker_config" }
-console_error_panic_hook = "0.1.1"
-console_log = { version = "1.0" }
-ed25519-dalek = { workspace = true, features = ["pkcs8"] }
-futures-util = "0.3.31"
-getrandom = { version = "0.2", features = ["js"] }
-hex = "0.4"
-log = { version = "0.4"}
-p256 = { version = "0.13", features = ["ecdsa"] }
-prometheus = "0.13.4"
+console_error_panic_hook.workspace = true
+console_log.workspace = true
+ed25519-dalek.workspace = true
+futures-util.workspace = true
+getrandom.workspace = true
+hex.workspace = true
+log.workspace = true
+p256.workspace = true
+prometheus.workspace = true
 rand.workspace = true
-serde_bytes = "0.11"
+serde-wasm-bindgen.workspace = true
+serde.workspace = true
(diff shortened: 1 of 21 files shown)


crates/static_ct_api/src/lib.rs
```diff
@@ -6,3 +6,58 @@ pub mod static_ct;
 
 pub use rfc6962::*;
 pub use static_ct::*;
+
+pub type Result<T> = core::result::Result<T, Error>;
```
cjpatton, line 10: In my opinion, the pattern of defining your own `Result` in a crate (with your own custom error) is a bad pattern. There often cases where you have the `std` Result and your own Result in the same module, and this hurts readability.

crates/static_ct_api/src/lib.rs
```diff
@@ -6,3 +6,58 @@ pub mod static_ct;
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
cjpatton, line 13: I would give your crate a distinguished name for errors, e.g., if the crate is called "azul", you might go with `AzulError`.

crates/static_ct_api/src/static_ct.rs
```diff
@@ -162,9 +162,13 @@ pub fn tile_path(t: &Tile) -> String {
 /// # Errors
 ///
 /// Returns an error if decoding the verifying key fails.
-pub fn log_id_from_key(vkey: &EcdsaVerifyingKey) -> Result<[u8; 32], x509_verify::spki::Error> {
-    let pkix = vkey.to_public_key_der()?;
-    Ok(Sha256::digest(&pkix).into())
+///
+/// # Panics
+///
+/// Panics if decoding the verifying key fails.
```
cjpatton, line 168: Why panic instead of err?

crates/ct_worker/Cargo.toml
```diff
@@ -49,25 +49,26 @@ config = { path = "./config", package = "ct_worker_config" }
 console_error_panic_hook = "0.1.1"
 console_log = { version = "1.0" }
 ed25519-dalek = { workspace = true, features = ["pkcs8"] }
+futures-util = "0.3.31"
```
cjpatton, line 52: Should this be a workspace dependency? Here and below

cjpatton (approved):
Beautiful! You've got documentation conventions down pat.

crates/static_ct_api/src/lib.rs
lukevalenta: Fair enough! I saw this pattern used in a few places, but wasn't sure what the best practice is. I did indeed have to `use std::result::Result as StdResult` in a couple places.

crates/ct_worker/Cargo.toml
lukevalenta: Hmm so I was just putting dependencies that are only used in a single crate within that crate's Cargo.toml, but seems that logs of people just put everything as workspace dependencies and call it a day: https://www.reddit.com/r/rust/comments/1i4c1x5/when_should_a_dependency_be_in_the_workspace_vs/. I'll do the same, since I think it'll make life easier!

crates/static_ct_api/src/static_ct.rs
lukevalenta: Eh it was just to avoid having to thread the error through to the application, since I don't really expect it to ever actually occur. Fixed now regardless!

crates/ct_worker/Cargo.toml
lukevalenta: Done!

lukevalenta:
> Beautiful! You've got documentation conventions down pat.

Heh, just a very strict linter :).

PR #36, opened 2025-05-14
Add TlogTile wrapper around Tile to implement tlog-tiles spec

lukevalenta:
The Tile struct supports the tile format used by the Go checksum library (from which it was ported), but the [tlog-tiles](https://c2sp.org/tlog-tiles) spec makes a few changes. Specifically, it fixes the height at 8 (so each tile has 256 elements), and uses the path element "entries" for data tiles.

With this change, we can remove custom path logic from the static_ct_api crate, although we still need to use the path element "data".

I don't love the "data_path_opt" field in tiles, but this seems like an OK way to support both `"data"` and `"entries"` as the data path element.

It will also allow us to support non-standard data tile paths like `"extra_data"`, if we so choose.

### DIFF
--- a/crates/tlog_tiles/src/tile.rs
+++ b/crates/tlog_tiles/src/tile.rs
@@ -19,6 +19,7 @@ use crate::tlog::{
     node_hash, split_stored_hash_index, stored_hash_index, sub_tree_index, Hash, HashReader,
     TlogError, HASH_SIZE,
 };
+use std::cmp::max;
 use std::collections::HashMap;
 use std::fmt;
 use std::str::FromStr;
@@ -204,11 +205,7 @@ impl Tile {
     }
 
     /// Path returns a tile coordinate path describing t.
-    pub fn path(&self) -> String {
-        self.path_internal(true)
-    }
-
-    fn path_internal(&self, with_height: bool) -> String {
+    pub fn path(&self, with_height: bool) -> String {
         let mut n = self.n;
         let h_str = if with_height {
             &format!("/{}", self.h)
@@ -255,8 +252,7 @@ impl Tile {
         Some(t)
     }
 
-    /// Parses a tile coordinate path. This currently supports only paths with
-    /// height element and with the data path element "data".
+    /// Parses a tile coordinate path.
     ///
     /// # Errors
     ///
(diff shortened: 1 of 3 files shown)


crates/tlog_tiles/src/tile.rs
```diff
@@ -41,18 +41,19 @@ const PATH_BASE: u64 = 1000;
...
-    h: u8,         // height of tile (1 ≤ H ≤ 30)
-    l: u8,         // level in tiling (0 ≤ L ≤ 63)
-    n: u64,        // number within level (0 ≤ N, unbounded)
-    w: u32,        // width of tile (1 ≤ W ≤ 2**H; 2**H is complete tile)
-    is_data: bool, // whether or not this is a data tile
+    h: u8,                               // height of tile (1 ≤ H ≤ 30)
+    l: u8,                               // level in tiling (0 ≤ L ≤ 63)
+    n: u64,                              // number within level (0 ≤ N, unbounded)
+    w: u32,                              // width of tile (1 ≤ W ≤ 2**H; 2**H is complete tile)
+    data_path_opt: Option<&'static str>, // whether or not this is a data tile, and the data path element to use for encoding
```
cjpatton, line 56: Could this be an enum instead of a string constant? The benefit would be that the compiler would tell us if we accidentally used an unknown variant.

```rust
// not sure about the best name
pub enum TilePath{
  Data,
}

// perhaps we also need this:
impl AsRef<str> for TilePath {
  fn as_ref(&self) -> &str {
     match self {
        Self::Data => "data",
     }
  }
}
```

```suggestion
    data_path_opt: Option<TilePath>, // whether or not this is a data tile, and the data path element to use for encoding
```

crates/tlog_tiles/src/tile.rs
```diff
@@ -354,6 +368,73 @@ impl Tile {
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl std::ops::DerefMut for TlogTile {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
cjpatton, line 388: Are you implementing Deref(Mut) because there are methods on `Tile` do we want `TlogTile` to have?

As far as I understand, this pattern is discouraged. I'm not sure why, but I've had trouble with this in the past. However,  there are cases where it makes sense and this is one of them.

cjpatton (approved):

PR #37, opened 2025-05-16
Upload staging bundle to fixed path in DO storage

### LINKED ISSUE #34: Write staging bundle to a single location
As part of the two-step commit design discussed in https://github.com/FiloSottile/sunlight/issues/11, the Sequencer uploads all tiles in a staging bundle to a unique location (tied to the tree size and root hash), updates lock backend, applies the staged tiles to the public bucket, and then finally updates the checkpoint in the public bucket.

Writing each staging bundle to a unique location is a little annoying, since then those bundles either waste space or require special cleanup later. What if we just encode the tree size and root hash into the staging bundle itself, and only write to a single staging file location (e.g., `/staging`)?

lukevalenta:
-  714c7a0    
    Add support for multipart uploads to DO storage backend

    The KV API for DO storage limits values to 2MB, which we could exceed
    when sequencing moderately-sized batches (1000 entries of size 2kB
    each). Add support for splitting these writes across multiple DO storage
    values to avoid this limit.

-  b92b8bc
    
    Upload staging bundle to fixed path in DO storage

    This commit includes two major changes:

    - Write staging bundle to a fixed path, fixes #34.

      This means that we don't need to clean up old staging bundles from
      object storage. Insert the current tree size and hash into the staging
      bundle for good measure, but I don't believe this is necessary.

      As an added bonus, this cuts out one R2 write/sec (replaced w/ a write
      to DO storage, which should be faster).

    - Write staging bundle to DO storage instead of R2

      The [reasons](https://github.com/FiloSottile/sunlight/issues/310) not
      to use a fixed path in sunlight are to avoid zombie writes/rollbacks,
      and against multiple sequencers writing at the same time. With the
      strong consistency guarantees provided by Durable Objects, these
      failure modes are not applicable.

      Zombie writes: The DO storage relay service requires that writes are
      [acknowledged](https://blog.cloudflare.com/sqlite-in-durable-objects/) by a least three of five "follower" machines, which helps
      to ensure that all confirmed writes are correctly persisted in the
      proper order.

      Multiple sequencers: Durable Objects also guarantees that only a
      single 'scheduled' task can run at a time, so we don't have to worry
      about multiple sequencers.

    Also reword and simplify the storage 'break' modes in the tests.

### DIFF
--- a/crates/ct_worker/src/lib.rs
+++ b/crates/ct_worker/src/lib.rs
@@ -18,6 +18,7 @@ use metrics::{millis_diff_as_secs, AsF64, ObjectMetrics};
 use p256::{ecdsa::SigningKey as EcdsaSigningKey, pkcs8::DecodePrivateKey};
 use serde::Deserialize;
 use serde_bytes::ByteBuf;
+use sha2::{Digest, Sha256};
 use static_ct_api::{LookupKey, UnixTimestamp};
 use std::collections::{HashMap, VecDeque};
 use std::io::Write;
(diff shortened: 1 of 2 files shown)


crates/ct_worker/src/lib.rs
```diff
@@ -279,21 +280,106 @@ impl MemoryCache {
 
 // Compare-and-swap backend.
 trait LockBackend {
-    async fn put(&self, key: &str, checkpoint: &[u8]) -> Result<()>;
+    const PART_SIZE: usize;
```
cjpatton, line 283: Worth documenting whether we mean bytes, kilobytes, or megabytes.

Also, this is probably a _maximum_ part size, correct? Since if `size % PART_SZIE > 0` then the last part won't have length `PART_SIZE`

crates/ct_worker/src/lib.rs
```diff
@@ -279,21 +280,106 @@ impl MemoryCache {
...
+    async fn get_multipart(&self, key: &str) -> Result<Vec<u8>>;
+    async fn put(&self, key: &str, value: &[u8]) -> Result<()>;
     async fn swap(&self, key: &str, old: &[u8], new: &[u8]) -> Result<()>;
     async fn get(&self, key: &str) -> Result<Vec<u8>>;
 }
 
 impl LockBackend for State {
+    // Stated limit is 2MB, but actual limit seems to be between 2.1 and 2.2MB.
+    // https://developers.cloudflare.com/durable-objects/platform/limits/
+    const PART_SIZE: usize = 2_000_000;
```
cjpatton, line 295: nit: Should this be a multiple of 1024 * 1024?

crates/ct_worker/src/lib.rs
```diff
@@ -279,21 +280,106 @@ impl MemoryCache {
...
+                .map_err(|_| "u32 conversion failed")?,
+        ) as usize;
+        let checksum: [u8; 32] = manifest[4..4 + 32]
+            .try_into()
+            .map_err(|_| "slice conversion failed")?;
+        if len > Self::PART_SIZE * Self::MAX_PARTS {
+            return Err("value too large".into());
+        }
+        let mut result = Vec::with_capacity(len);
+        let keys = (0..((len - 1 + Self::PART_SIZE) / Self::PART_SIZE))
```
cjpatton, line 352: Check out https://doc.rust-lang.org/std/primitive.u32.html#method.div_ceil

cjpatton (approved):

crates/ct_worker/src/lib.rs
lukevalenta: Much nicer, thanks!

Issue #33, opened 2025-05-12
Prefer publishing full tiles


lukevalenta:
Add ability to delay submissions for up to a few seconds to try to publish full data tiles instead of partial tiles whenever possible.

PR #38, opened 2025-05-16
Refactor pool logic

lukevalenta:
This PR contains quite a bit of refactoring of the pool logic in order to address #33. The most significant change is using a separate https://docs.rs/tokio/latest/tokio/sync/watch/index.html `{Sender, Receiver}` pair per entry instead of per "pool" of entries. This gives us the flexibility to break up pools of pending entries so we can hold back some entries when sequencing to try to prefer creating full tiles.

The commits can be reviewed independently and each have more detailed commit messages, although some of the commits remove or refactor code added in previous commits (e.g., the `Pool` struct).

* 2af4060 Add ability to hold entries back from sequencing to reduce partial tiles
* f9cc093 Use Vec::with_capacity where applicable, and pass owned PendingLogEntry to avoid clone
* 3a0f655 Use one Sender per PendingLogEntry instead of per batch
* 41ec703 Add PendingLogEntry struct to avoid overloading LogEntry
* 7fce31e Merge Pool struct into PoolState
* 143fd22 Add pool receiver directly to in_sequencing HashMap
* 17261a8 Refactor pool logic

### DIFF
--- a/crates/ct_worker/src/ctlog.rs
+++ b/crates/ct_worker/src/ctlog.rs
@@ -67,7 +67,7 @@ pub(crate) struct LogConfig {
 /// Ephemeral state for pooling entries to the CT log.
 ///
 /// The pool is written to by `add_leaf_to_pool`, and by the sequencer
-/// when rotating `pending` and `in_sequencing`.
+/// when rotating pending and in-sequencing entries.
 ///
 /// As long as the above-mentioned blocks run synchronously (no 'await's), Durable Objects'
 /// single-threaded execution guarantees that `add_leaf_to_pool` will never add to a pool that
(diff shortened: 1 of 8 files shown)


crates/ct_worker/src/ctlog.rs
```diff
@@ -70,12 +76,110 @@ pub(crate) struct LogConfig {
...
 #[derive(Default, Debug)]
 pub(crate) struct PoolState {
-    current_pool: Pool,
-    // in_sequencing is the [Pool::by_hash] map of the pool that's currently being
-    // sequenced. These entries might not be sequenced yet or might not yet be
-    // committed to the deduplication cache.
-    in_sequencing: HashMap<LookupKey, u64>,
-    in_sequencing_done: Option<Receiver<SequenceMetadata>>,
+    // How many times the oldest entry has been held back from sequencing.
+    holds: usize,
```
rozbb, line 80: Nit: I got confused reading the function below bc I thought this was "number of entries being held back". Maybe rename `oldest_pending_entry_holds` or something (doesn't get referenced a lot, so maybe the verbosity is okay)

crates/ct_worker/src/ctlog.rs
```diff
@@ -119,9 +123,57 @@ impl PoolState {
...
+        if publishing_full_tile {
+            // We're going to publish at least one full tile which will contain
+            // any leftover entries from the previous sequencing. Reset the
+            // count since the new leftover entries have not yet been held back.
+            self.holds = 0;
+        }
+        // Flush all of the leftover entries if the oldest is before the cutoff.
+        let flush_oldest = self.holds >= max_pending_entry_holds;
+
+        if leftover == 0 || flush_oldest {
```
rozbb, line 148: So if `flush_oldest` is triggered when there's 1.5 tiles pending, we'll publish 1.5 files? Why not just publish a full one and hold back the half?

crates/ct_worker/src/ctlog.rs
```diff
@@ -98,10 +75,17 @@ impl Default for Pool {
...
 #[derive(Default, Debug)]
 pub(crate) struct PoolState {
-    current_pool: Pool,
-    // in_sequencing is the [Pool::by_hash] map of the pool that's currently being
-    // sequenced. These entries might not be sequenced yet or might not yet be
-    // committed to the deduplication cache.
+    // Entries that are ready to be sequenced.
+    pending_entries: Vec<LogEntry>,
+
+    // Deduplication cache for entries currently pending sequencing.
```
rozbb: Is this a dedup cache for `pending_entries` then? Or something else

rozbb (approved):
Looks good! Just some small notes

crates/ct_worker/src/ctlog.rs
lukevalenta: Yes, exactly! I'll rename for clarity.

crates/ct_worker/src/ctlog.rs
lukevalenta: Good idea. Done!

PR #43, opened 2025-05-22
Add ability to tune per-log batching behavior

lukevalenta:
c913c38 Fix comment about dropped channels

The value in the channel will persist even after the sender is dropped,
so remove the comment.  According to
https://docs.rs/tokio/latest/tokio/sync/watch/index.html#closing, "The
value in the channel will not be dropped until the sender and all
receivers have been dropped."

f721b0c Add breakdown of full vs partial tiles to metrics

a053648 Add ability to set a maximum timeout for skipped entries

When a pending entry has been queued for longer than the timeout, it
will be included in the next sequencing regardless of how many times it
has exceeded the number of allowed skipped sequencings.

859df92 Add option to disable deduplication checks

Add an option to the configuration to disable the deduplication cache,
which is useful for running tests and benchmarks against a log.

a16ee94 Add batcher parameters to log configuration, and allow setting zero batchers

- Add ability to set num_batchers to zero so the frontend worker
  connects directly to the sequencer.
- Remove MAX_IN_FLIGHT check from batcher. If we need some sort of
  rate-limiting in the batchers, we can add it later based on
  operational experience.

0f92d2a Allow skipping checkpoint creation when there are no entries to sequence

- Restore old log behavior by setting default 'max_sequence_skips' to 0
- Rework PoolState::take() to return an Option indicating whether or not
  we sequencing can be skipped are no new entries or entries that don't
  make up a full tile.
- Set additionalProperties:false in JSON schema

e9d76c0 Specify sequence interval in milliseconds instead of seconds

### DIFF
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -77,7 +77,12 @@
                         "max_sequence_skips": {
                             "type": "integer",
                             "default": 0,
-                            "description": "The maximum number of times sequencing can be skipped to avoid creating partial tiles. If non-zero, pending entries may be delayed by a multiple of the sequence interval."
+                            "description": "The maximum number of times sequencing can be skipped to avoid creating partial tiles. If non-zero, pending entries may be delayed by either a multiple of the sequence interval or sequence_skip_threshold_millis if set."
+                        },
+                        "sequence_skip_threshold_millis": {
+                            "type": "integer",
+                            "default": 0,
+                            "description": "If non-zero, entries will only be skipped by sequencing (when max_sequenced_skips is non-zero) if they have been in the pool for less than this timeout."
                         },
                         "num_batchers": {
                             "type": "integer",
@@ -90,7 +95,7 @@
                             "type": "integer",
                             "minimum": 100,
                             "default": 1000,
-                            "description": "The maximum time to wait before submitting a batch to the sequencer, in milliseconds."
+                            "description": "The maximum duration to wait before submitting a batch to the sequencer, in milliseconds."
                         },
                         "max_batch_entries": {
                             "type": "integer",
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -76,6 +76,7 @@
                         },
                         "max_sequence_skips": {
                             "type": "integer",
+                            "minimum": 0,
                             "default": 0,
                             "description": "The maximum number of times sequencing can be skipped to avoid creating partial tiles. If non-zero, pending entries may be delayed by either a multiple of the sequence interval or sequence_skip_threshold_millis if set."
                         },
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -106,7 +106,7 @@
                         "enable_dedup": {
                             "type": "boolean",
                             "default": true,
-                            "description": "Enables checking the deduplication cache for add-(pre-)chain requests. Can be disable for tests and benchmarks."
+                            "description": "Enables checking the deduplication cache for add-(pre-)chain requests. Can be disabled for tests and benchmarks."
                         }
                     },
                     "required": [
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -78,6 +78,25 @@
                             "type": "integer",
                             "default": 0,
                             "description": "The maximum number of times sequencing can be skipped to avoid creating partial tiles. If non-zero, pending entries may be delayed by a multiple of the sequence interval."
+                        },
+                        "num_batchers": {
+                            "type": "integer",
+                            "minimum": 0,
+                            "default": 8,
+                            "maximum": 255,
+                            "description": "The number of batchers to use to proxy requests to the sequencer. If zero, requests from the frontend worker go directly to the sequencer."
+                        },
+                        "batch_timeout_millis": {
+                            "type": "integer",
+                            "minimum": 100,
+                            "default": 1000,
+                            "description": "The maximum time to wait before submitting a batch to the sequencer, in milliseconds."
+                        },
+                        "max_batch_entries": {
+                            "type": "integer",
+                            "minimum": 1,
+                            "default": 1000,
+                            "description": "The maximum number of entries per batch."
                         }
                     },
                     "required": [
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -97,6 +97,11 @@
                             "minimum": 1,
                             "default": 1000,
                             "description": "The maximum number of entries per batch."
+                        },
+                        "disable_dedup": {
+                            "type": "boolean",
+                            "default": false,
+                            "description": "Disable checking the deduplication cache for add-(pre-)chain requests (e.g., for tests or benchmarks)."
                         }
                     },
                     "required": [
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -1,6 +1,7 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
     "type": "object",
+    "additionalProperties": false,
     "properties": {
         "logging_level": {
             "type": "string",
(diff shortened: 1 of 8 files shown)


crates/ct_worker/src/ctlog.rs
```diff
@@ -161,19 +176,27 @@ impl<E: PendingLogEntryTrait> PoolState<E> {
         &mut self,
```
cjpatton, line 176: ^ just double checking if the comment for this function needs to be updated.

crates/ct_worker/src/ctlog.rs
```diff
@@ -161,19 +176,27 @@ impl<E: PendingLogEntryTrait> PoolState<E> {
...
         max_sequence_skips: usize,
+        sequence_skip_threshold_millis: Option<u64>,
     ) -> Option<Vec<(E, Sender<SequenceMetadata>)>> {
         let new_size = old_size + self.pending_entries.len() as u64;
         let publishing_full_tile =
             new_size / u64::from(TlogTile::FULL_WIDTH) > old_size / u64::from(TlogTile::FULL_WIDTH);
-
-        if publishing_full_tile && max_sequence_skips > 0 {
+        let num_leftover_entries =
+            usize::try_from(new_size % u64::from(TlogTile::FULL_WIDTH)).unwrap();
```
cjpatton, line 185: What do we mean by "leftover"? What happens if `new_size > FULL_WIDTH`?

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -260,10 +259,25 @@ async fn add_chain_or_pre_chain(
...
+        if kv
+            .put(&BASE64_STANDARD.encode(lookup_key), "")
+            .unwrap()
+            .metadata::<SequenceMetadata>(metadata)
+            .unwrap()
+            .execute()
+            .await
+            .is_err()
+        {
+            debug!("{name}: Failed to write entry to deduplication cache");
```
cjpatton, line 280: Should this be a `warn!` or an `error!`?

crates/ct_worker/config.schema.json
```diff
@@ -70,10 +74,10 @@
                             "default": 1000,
                             "description": "The duration in between sequencing operations, in milliseconds."
                         },
-                        "max_pending_entry_holds": {
+                        "max_sequence_skips": {
                             "type": "integer",
-                            "default": 1,
-                            "description": "The maximum number of times a pending entry can be held back from sequencing to avoid creating partial tiles. If non-zero, pending entries may be delayed by a multiple of sequence interval."
+                            "default": 0,
+                            "description": "The maximum number of times sequencing can be skipped to avoid creating partial tiles. If non-zero, pending entries may be delayed by a multiple of the sequence interval."
```
cjpatton: Observation: Lots of changes to the schema in this PR. Whenever we make changes, we need to sync it up with the code, e.g., for enforcing bounds on integers. Are you worried about this at all?

crates/ct_worker/src/batcher_do.rs
```diff
@@ -18,26 +21,17 @@ use tokio::sync::watch::{self, Sender};
...
-const MAX_BATCH_SIZE: usize = 100;
-
-// The maximum amount of time to wait before submitting a batch.
-const MAX_BATCH_TIMEOUT_MILLIS: u64 = 1_000;
-
 #[durable_object]
 struct Batcher {
     env: Env,
-    batch: Batch,
+    pending_batch: Batch,
```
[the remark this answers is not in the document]
lukevalenta: The batcher aggregates entries and then submits them in a batch to the sequencer, and the 'pending' batch is the one currently getting filled up.

crates/ct_worker/src/ctlog.rs
lukevalenta: In a tiled log, each full tile contains `2^TILE_HEIGHT = FULL_WIDTH` entries, and a partial tile is one (always on the right 'edge' of the tree) that is still getting filled up. In static CT, full tiles contain 256 entries. So if you have a tiled log of size 258, that means you have one full level-0 tile with 256 entries and one partial level-0 tile with 2 entries. If you grow the log by, say, 5 entries, you'll then have new partial tile with 7 entries that you have to publish to replace the old partial tile. (Full tiles never change.)

What this change is doing is trying to delay producing partial tiles when they're just going to be replaced. So if we have a tree of size 258 and then add 100 entries and then add 300 more, we'll skip producing a partial tile for (0-indexed) entries 256-357 since that's about to be replaced by the the full tile for entries 256-511. After the 100 entries are added, `num_leftover_entries` would be 358 % 256 = 102.

crates/ct_worker/config.schema.json
lukevalenta: The schema is statically checked in `build.rs`, so the code will fail to compile if the provided config doesn't match the schema, including the bounds checks. For example, if I set minimum `batch_timeout_millis` to 101 and then try to set the value to `100` in `config.dev.json`, I get this error:
```
% cargo build
   Compiling ct_worker v0.2.0 (/Users/lvalenta/cf-repos/azul/crates/ct_worker)
error: failed to run custom build command for `ct_worker v0.2.0 (/Users/lvalenta/cf-repos/azul/crates/ct_worker)`

Caused by:
  process didn't exit successfully: `/Users/lvalenta/cf-repos/azul/target/debug/build/ct_worker-1e3be4eee95bc251/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' panicked at crates/ct_worker/build.rs:29:9:
  config 'config.dev.json' does not match schema 'config.schema.json': 100 is less than the minimum of 101
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```

The main thing to sync in the code is setting the default values in `config/src/lib.rs`, which I think is right but there is room for human error.

File: crates/ct_worker/src/metrics.rs [hunk 6]
```diff
@@ -114,9 +114,10 @@ impl Metrics {
             r
         )
         .unwrap();
-        let seq_data_tile_size = register_histogram_with_registry!(
+        let seq_data_tile_size = register_histogram_vec_with_registry!(
             "sequencing_data_tiles_bytes",
             "Size of uploaded data tiles, including partials.",
+            &["type"],
```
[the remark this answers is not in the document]
lukevalenta: These are the only two places where the labels are used and it's pretty low stakes if we have a typo (it's just the metrics label), so I'm kind of inclined to leave as-is for now, but will keep the suggestion in mind for the future!

crates/ct_worker/config.schema.json
cjpatton: That is EXCELLENT!

crates/ct_worker/src/ctlog.rs
cjpatton: Thanks for the explanation.

cjpatton (approved):

crates/ct_worker/src/ctlog.rs
```diff
@@ -147,61 +147,57 @@ impl<E: PendingLogEntryTrait> PoolState<E> {
...
-            } else {
-                // We didn't fill up a full tile, so nothing to return.
-                Vec::new()
-            }
+            // Skip this checkpoint. There are no full tiles to sequence, and
+            // we're below the skip threshold.
+            self.sequence_skips += 1;
+            None
         }
     }
```
rozbb, line 228: I feel like I can follow the logic of this function a lot better now. Nice!

crates/ct_worker/config.schema.json
```diff
@@ -97,6 +97,11 @@
                             "minimum": 1,
                             "default": 1000,
                             "description": "The maximum number of entries per batch."
+                        },
+                        "disable_dedup": {
```
rozbb, line 106: Nit: I think it's generally better to have the flag be the positive form. So `enable_dedup` with default true. "Enables checking the deduplication cache for add-(pre-)chain requests. Should be disabled, e.g., for tests and benchmarks."

rozbb (approved):
nice!

crates/ct_worker/config.schema.json
lukevalenta: That's fair! I was on the fence for this one but your comment is the deciding factor. Updated!

PR #44, opened 2025-06-02
Make Sequencer and Batcher generic over the type of log being stored

rozbb:
Sorry, this is a big PR. At the highest level, `Sequencer` was doing a lot of CT-specific actions: loading RFC6962 signing keys, logging cert info, hashing leaves in a specific way, signing precerts in a specific way, etc. This PR (hopefully) has genericized all of these operations, so that `Sequencer` can be instantiated over many kinds of logs. The CT functionality was moved to `StaticCTSequencer` and related log entry types.

I think each commit is pretty coherent within itself, so I recommend going thru commit-by-commit. And ofc ask me any questions.


lukevalenta:
Could you rebase on the latest main branch? I think you can just drop commits 17261a8..2af4060 since those are already merged into main. (I think you must have forked off of one of the not-yet-merged PRs, but then all the commit hashes changed when the PR was "Rebase and merge"d.)

crates/static_ct_api/src/static_ct.rs
```diff
@@ -131,39 +149,50 @@ pub fn log_id_from_key(vkey: &EcdsaVerifyingKey) -> Result<[u8; 32], StaticCTErr
...
+pub trait LogEntryTrait: core::fmt::Debug + Sized {
+    /// The pending version of this log entry. Usually the same thing but doesn't have a timestamp or tree index
+    type Pending: PendingLogEntryTrait;
+
+    /// The error type for [`Self::parse_from_tile_entry`]
+    type ParseError: std::error::Error + Send + Sync + 'static;
+
+    fn new(pending: Self::Pending, timestamp: UnixTimestamp, leaf_index: u64) -> Self;
+
+    /// Returns the underlying pending entry
```
lukevalenta, line 175: minor nit: Rust style guide says comments should end with a period: https://doc.rust-lang.org/nightly/style-guide/#comments.

crates/static_ct_api/src/static_ct.rs
```diff
@@ -131,39 +149,50 @@ pub fn log_id_from_key(vkey: &EcdsaVerifyingKey) -> Result<[u8; 32], StaticCTErr
...
 
-    /// The `PrecertChainEntry.pre_certificate`.
-    /// It must be at most 2^24-1 bytes long.
-    pub pre_certificate: Vec<u8>,
+    /// The canonical byte representation of this object. Only used by [`GenericLogEntry`].
+    fn as_bytes(&self) -> &[u8];
 }
 
-pub type LookupKey = [u8; 16];
+pub trait LogEntryTrait: core::fmt::Debug + Sized {
```
lukevalenta, line 166: Is the plan to move these traits (`LogEntryTrait` and `PendingLogEntryTrait`) to a separate crate from `static_ct_api` crate?

When we do that, let's maybe rename these to `SequencedEntryTrait` and `PendingEntryTrait`?

crates/static_ct_api/src/static_ct.rs
```diff
@@ -365,6 +415,90 @@ impl TileIterator {
     }
 }
 
+/* TODO: Uncomment this once the static CT types are implemented
+/// A generic log entry compatible with the [tlog-tiles spec](https://github.com/C2SP/C2SP/blob/main/tlog-tiles.md)
+#[derive(Debug)]
+struct GenericLogEntry<E: PendingLogEntryTrait>(E);
+
+impl<E: PendingLogEntryTrait> LogEntryTrait<E> for GenericLogEntry<E> {
+    fn new(pending: E, _: UnixTimestamp, _: u64) -> Self {
```
lukevalenta, line 424: One of the refactors I made in https://github.com/cloudflare/azul/pull/40 was defining a `SequenceMetadata` alias for `(u64, u64)` to wrap the index and timestamp so we aren't passing opaque `u64`s around. It seemed to make things much cleaner, if you want to apply it to your PR as well.

crates/static_ct_api/src/static_ct.rs
```diff
@@ -421,87 +555,75 @@ impl TreeWithTimestamp {
...
     fn key_id(&self) -> u32 {
         self.v.key_id()
     }
-    fn sign(&self, msg: &[u8]) -> Result<Vec<u8>, signature::Error> {
-        let sig = self.k.try_sign(msg)?;
-        Ok(sig.to_vec())
+
+    fn sign(
+        &self,
+        _timestamp_unix_millis: u64,
```
lukevalenta, line 611:
```suggestion
        _: UnixTimestamp,
```

crates/signed_note/src/lib.rs
```diff
@@ -269,10 +273,6 @@ pub trait Signer {
 
 /// Computes the key ID for the given server name and encoded public key
 /// as RECOMMENDED at <https://c2sp.org/signed-note#signatures>.
-///
-/// # Panics
```
lukevalenta, line 273: I think I had this comment in here to avoid a linter complaint (with pedantic), but I can fix later.

crates/signed_note/src/lib.rs
```diff
@@ -248,6 +248,10 @@ pub trait Verifier {
 
     /// Reports whether sig is a valid signature of msg.
     fn verify(&self, msg: &[u8], sig: &[u8]) -> bool;
+
+    /// Extracts a Unix timestamp in milliseconds from the given signature bytes, if defined. Errors if
+    /// the signature is malformed.
+    fn extract_timestamp_millis(&self, sig: &[u8]) -> Result<Option<u64>, ()>;
```
lukevalenta, line 254: Nice addition to the trait! If we define the `UnixTimestamp` alias in some shared crate later we can use it here.
```suggestion
    fn extract_timestamp_millis(&self, sig: &[u8]) -> Result<Option<UnixTimestamp>, ()>;
```

crates/signed_note/src/lib.rs
```diff
@@ -259,7 +263,7 @@ pub trait Signer {
     /// Returns the key ID.
     fn key_id(&self) -> u32;
 
-    /// Returns a signature for the given message.
+    /// Returns a signature for the Note. Uses the origin returns by `self.name()`.
```
lukevalenta, line 266:
```suggestion
    /// Returns a signature for the Note. Uses the origin returned by `self.name()`.
```

lukevalenta (commented):
Only got partway through, but I'll leave the comments so far.

rozbb:
> Could you rebase on the latest main branch? You probably forked off of one of the not-yet-merged PRs, but then all the commit hashes changed when the PR was "Rebase and merge"-ed.
> 
> I _think_ it should work if you `git rebase -i main` from your branch and then just delete the duplicated commits [17261a8](https://github.com/cloudflare/azul/commit/17261a8a245eda0d1a84309ad9ddfd6f7f15c420)..[2af4060](https://github.com/cloudflare/azul/commit/2af4060b02be9a06e96ea7c245886549ac62bc4e).

I just fought git for nearly an hour and got nowhere :( Do you wanna try, or maybe just squash merge?

crates/ct_worker/src/ctlog.rs
```diff
@@ -1393,27 +1431,63 @@ mod tests {
...
+            &log.config,
+            &log.object,
+            &log.lock,
+        ))
+        .unwrap_err();
     }
 
     #[test]
     fn test_reload_wrong_key() {
+        // Need to use a seed because we can't clone a log config
```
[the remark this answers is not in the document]
lukevalenta: I removed this comment since it wasn't making sense to me, but please push back if it's still relevant!

lukevalenta:
very minor nit (no need to update the PR): the de-facto norm is imperative verb form for git messages, so a commit message would be "Rename x to y" instead of "Renamed x to y", so you could say "If this commit is applied, it will <commit-message>": https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53#commit-message-guidelines

crates/static_ct_api/src/static_ct.rs
rozbb: Yeah that's exactly the plan. And agreed on naming. Tho it'd be nice to name a trait something without "trait" in the name. We'll see

crates/tlog_tiles/src/checkpoint.rs
```diff
@@ -271,6 +272,33 @@ impl Checkpoint {
     }
 }
 
+/// An object that can produce a [note signature](https://github.com/C2SP/C2SP/blob/main/signed-note.md) for a given checkpoint
+pub trait CheckpointSigner {
```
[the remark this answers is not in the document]
rozbb: Good q, I should have put my reasoning at the top. A plain NoteSigner is not sufficient because it only provides the implementer a raw `msg: &[u8]`. This is insufficient for a RFC6962 signing process, since it needs to know the timestamp, tree size, and root hash. The workaround for this rn is to separately create the signature using these values gotten from elsewhere, then make a dummy signer (`InjectedSigner`) that's prepopulated with the signature.

The method I'm proposing is to make a signing trait that exposes the checkpoint components to the signer. This is compatible with RFC6962 signers, tlog-cosignature signers, and plain Ed25519 signers. It also does away with `InjectedSigner`.

crates/ct_worker/src/sequencer_do.rs
```diff
@@ -239,9 +281,10 @@ impl Sequencer {
                 pending_entry,
             );
 
+            logging_labels.push(add_leaf_result.source().to_string());
             self.metrics
                 .entry_count
-                .with_label_values(&[typ, add_leaf_result.source()])
+                .with_label_values(&logging_labels)
```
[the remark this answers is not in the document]
rozbb: I'm not sure I understand. What's the issue that arises if `Foo::logging_labels()` returns a vec of size 3 and `Bar::logging_labels()` returns a vec of size 8?

crates/ct_worker/src/ctlog.rs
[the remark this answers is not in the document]
rozbb: Ah that was an old comment from an intermediate solution that didn't work out. Ty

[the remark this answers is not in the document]
rozbb: I ended up making `clone_without_signers( )` to solve this. It's because `Box<dyn CheckpointSigner>` is not `Clone` and cannot be made `Clone` because `clone(&self) -> Self` is not a dyn-compatible method

rozbb:
> very minor nit (no need to update the PR): the de-facto norm is imperative verb form for git messages, so a commit message would be "Rename x to y" instead of "Renamed x to y", so you could say "If this commit is applied, it will "

Sorry, ya this is an old bad habit of mine

crates/ct_worker/src/ctlog.rs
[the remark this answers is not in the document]
lukevalenta: Nice, seems like a reasonable workaround!

crates/tlog_tiles/src/checkpoint.rs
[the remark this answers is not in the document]
lukevalenta: Ah I hadn't registered the extra parameters to `sign`. That makes sense!

PR #54, opened 2025-06-24
Add log.v3.json endpoint and clean up handler, fixes #41

### LINKED ISSUE #41: Add log.v3.json endpoint
Modify metadata endpoint to be compatible with log.v3.json log spec, (e.g., https://groups.google.com/a/chromium.org/g/ct-policy/c/htn88pupAqI/m/Hdbf7tBPAAAJ).

lukevalenta:
- Add an outer Router in the fetch handler to check the log parameter, instead of requiring each handler to check that the requested log path is in the allowlist
- Change /metadata endpoint to /log.v3.json, dropping the witness_key field. The witness is just used to exercise the code path, and it's not expected that any clients would actually use the endpoint.
- Disable bucket access via the Worker if monitoring_url is specified so clients must use the intended monitoring API endpoint.
- Update MMD to 60 as that's the max allowed in Chrome's CT program for static CT logs. (Note that this is effectively zero in the current implementation.)

### DIFF
--- a/crates/ct_worker/src/frontend_worker.rs
+++ b/crates/ct_worker/src/frontend_worker.rs
@@ -5,7 +5,7 @@
 
 use std::sync::LazyLock;
 
-use crate::{load_signing_key, load_witness_key, LookupKey, SequenceMetadata, CONFIG, ROOTS};
+use crate::{load_signing_key, LookupKey, SequenceMetadata, CONFIG, ROOTS};
 use config::TemporalInterval;
 use futures_util::future::try_join_all;
 use generic_log_worker::{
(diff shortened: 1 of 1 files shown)


crates/ct_worker/src/frontend_worker.rs
```diff
@@ -24,17 +24,17 @@ use tlog_tiles::{LogEntry, PendingLogEntry};
...
-// between when a SCT is issued and the corresponding entry is sequenced
-// in the log. For static CT logs, this is effectively zero since SCT issuance
-// happens only once the entry is sequenced. However, we can leave this value
-// in the metadata as the default (1 day).
-const MAX_MERGE_DELAY: usize = 86_400;
+// between when a SCT is issued and the corresponding entry is sequenced in the
+// log. For Azul-based logs, this is effectively zero since SCT issuance happens
+// only once the entry is sequenced. However, we can leave this value in the
+// metadata as the maximum allowed in Chrome's policy, 60 seconds.
+const MAX_MERGE_DELAY_SECS: usize = 60;
```
bwesterb, line 31: Why not zero, though?

lukevalenta: Mostly just to give flexibility in the (unlikely) event that we wanted to change the architecture to allow for a small merge delay (e.g., sequence entry and persist durably, return the SCT, and _then_ make the tiles public). The Tuscolo shards use 60 too: https://tuscolo2025h2.sunlight.geomys.org/log.v3.json.

PR #59, opened 2025-07-16
Add partial tile cleanup utility, fixes #1

### LINKED ISSUE #1: Clean up old staging and partial tiles from object backend
Add instructions for setting up a [cron trigger](https://developers.cloudflare.com/workers/configuration/cron-triggers/) or [object lifecycle rule](https://developers.cloudflare.com/r2/buckets/object-lifecycles/) to clean up old staging bundles and partial tiles where the corresponding full tile exists (e.g, `tile/data/001.p` can be deleted when `tile/data/001` exists).

lukevalenta:
The tlog-tiles and static-ct-api specs allow partial tiles to be deleted
when the corresponding full tile is available. This helps to reduce R2
storage costs, but will incur extra cost for the R2 list and delete
operations.

- Add a `Cleaner` Durable Object, which iterates over a log and for each
  full tile available in the public bucket, lists and deletes any
  corresponding partial tiles. The cleaner tracks subrequests to prevent
  an alarm from exceeding the 1000 subrequests limit per invocation.
- (bonus) Lays the groundwork for implementing a tlog-witness or
  tlog-mirror as a service that periodically updates based on a target
  log's latest checkpoint.

Other changes:
- Update to worker 0.6.1 for R2 `delete_multiple` support and
  `PartialEq` support for  ObjectIds.
- Move more Durable Object intialization logic from the MTC and CT
  applications to the `generic_log_worker` crate to deduplicate code.
- Fix tree time metric to seconds.

### DIFF
--- a/crates/ct_worker/config.schema.json
+++ b/crates/ct_worker/config.schema.json
@@ -112,7 +112,7 @@
                         "enable_ccadb_roots": {
                             "type": "boolean",
                             "default": true,
-                            "description": "Enables loading root store trusted roots from the CCADB list, in addition to any roots configured in `roots.<env>.pem`. If enabled, requires a KV namespace with the binding `ccadb_roots` to be configured in `wrangler.jsonc`, as well as a cron trigger so that the CCADB list auto-updates."
+                            "description": "Enables loading root store trusted roots from the CCADB list, in addition to any roots configured in `roots.<env>.pem`. If enabled, requires a KV namespace with the binding `ccadb_roots` to be configured in `wrangler.jsonc`, as well as a cron trigger so that the CCADB list auto-updates. The CCADB cron job only ever appends roots to the list, but manual changes can be made if necessary."
                         },
                         "clean_interval_secs": {
                             "type": "integer",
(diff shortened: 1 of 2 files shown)


crates/ct_worker/config.schema.json
```diff
@@ -113,6 +113,12 @@
                             "type": "boolean",
                             "default": true,
                             "description": "Enables loading root store trusted roots from the CCADB list, in addition to any roots configured in `roots.<env>.pem`. If enabled, requires a KV namespace with the binding `ccadb_roots` to be configured in `wrangler.jsonc`, as well as a cron trigger so that the CCADB list auto-updates."
+                        },
+                        "clean_interval_secs": {
+                            "type": "integer",
+                            "minimum": 1,
+                            "default": 60,
+                            "description": "How long to wait in between runs of the partial tile cleaner. For static CT, the cleaner can clean 498 tiles (127,488 entries) per run before hitting the Workers limit of 1000 subrequests, so the default of once every 60 seconds should keep up with a log that grows at 2000 entries/second."
```
bwesterb, line 121: How will we figure out if this is misconfigured / not keeping up?

bwesterb (approved):

crates/ct_worker/config.schema.json
lukevalenta: The cleaner prints logs every time it runs that include the 'cleaned size' (current progress) along with the largest known checkpoint size (which could be stale), so looking at the logs it's clear if the cleaner is keeping up.

It would probably be good to convert these to prometheus metrics in the future so it can be part of regular operational monitoring (but worst case, we just have extra files in R2 so it's not a big deal). I filed #75 to track.

PR #62, opened 2025-07-17
Update to worker 0.6.0, addressing breaking changes

lukevalenta:
- Durable Object methods no longer take a mutable reference, so we need to use interior mutability (RefCell) to update state, as recommended by the workers-rs maintainers. The tricky part is avoiding passing a borrowed Ref to an async function, as that could cause a panic if, for example, a concurrent async operation attempts to borrow an already-mutably-borrowed value.
- To simplify the above change, remove the Option wrapper from the GenericSequencer's 'sequence_state' field. Instead of loading the sequencer state just-in-time for the first sequencing, load it during the DO initialization (which happens after the first fetch or alarm). Similarly, when there is a fatal sequencing error requiring reloading the sequence state, load the state right away rather than waiting for the next sequencing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,9 +208,9 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


lukevalenta:
@rozbb fyi

bwesterb (approved):

PR #63, opened 2025-07-17
Make inclusion proof not depend on sequencer, fixes #61

### LINKED ISSUE #61: Make inclusion proof endpoint not depend on sequencer

lukevalenta:
- Remove sequencer as dependency for inclusion proofs, as all that is needed is the log config and R2 bucket.
- Add load_origin and load_checkpoint_signer helper functions.

We can extend this functionality to generate subtree inclusion proofs with https://github.com/cloudflare/azul/issues/64.


lukevalenta:
@rozbb fyi on this one as well

PR #72, opened 2025-07-24
Add ability to automatically fetch roots from CCADB, fixes #65

### LINKED ISSUE #65: Add ability to automatically fetch roots from CCADB
Similar to [sunlights implementation](https://github.com/FiloSottile/sunlight/blob/main/cmd/sunlight/roots.go), add the ability to automatically fetch roots from CCADB.

Currently, accepted roots are embedded in the worker binary. Two approaches to automatically add new roots from CCADB are as follows:

1) Run a script offline to manually regenerate the roots and then manually redeploy the worker with the new roots.
2) (Re-)add the ability for the log to load roots from Workers KV, and have a cron job periodically check for CCADB updates. This approach has more moving parts, but seems like the better long-term plan to reduce maintenance burden.

lukevalenta:
- Add a cron job that once a day fetches roots from CCADB that are included in root stores and saves them into per-log root files (only appending certs for each log shard, but never removing them). This is similar to the sunlight implementation.
- Add config flag (enabled by default) to specify if a log will attempt to load its roots from the CCADB roots file in Workers KV.
- Remove roots.default.pem and require per-environment roots files to exist (although they can be empty). These can be used to add additional accepted roots on top of the CCADB list, for instance, to allow test/staging CAs to submit log entries.
- Add more functionality to x509util::CertPool to support adding and checking inclusion of certs in an existing CertPool.


crates/ct_worker/README.md
```diff
@@ -150,7 +150,7 @@ Follow these instructions to deploy to a custom domain, suitable for running a p
 
 1.  Create a file `config.${ENV}.json` with the configuration for the log shards.
 
-1.  (Optional) Create a file `roots.${ENV}.pem` with any custom accepted roots for the log shards. By default, `roots.default.pem` will be used. All logs shards deployed within the same Worker script use the same set of roots. Roots can be updated later.
+1.  (Optional) Create a file `roots.${ENV}.pem` with any custom accepted roots for the log shards, in PEM format. If `enable_ccadb_roots` is set to true, these roots are used in addition to roots auto-pulled from the CCADB list. If `enable_ccadb_roots` is set to false for any logs in the deployment, `roots.${ENV}.pem` is required to exist and contain at least one certificate. All logs shards deployed within the same Worker script use the same set of additional roots. Roots can be updated later, but roots should generally not be removed once added.
```
bwesterb, line 153: What if a root is added to CCADB, then removed, and then the server restarts?

crates/ct_worker/src/ccadb_roots_cron.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+use x509_cert::{
+    der::{DecodePem, EncodePem},
+    Certificate,
+};
+use x509_util::CertPool;
+
+use crate::{ccadb_roots_filename, CCADB_ROOTS_NAMESPACE, CONFIG};
+
+#[event(scheduled)]
```
bwesterb, line 21: Can this function be called while a previous call is still running?

crates/ct_worker/src/ccadb_roots_cron.rs
```diff
@@ -0,0 +1,150 @@
...
+    };
+    if let Err(e) = update_ccadb_roots(&keys, &kv).await {
+        log::warn!("Failed to update CCADB roots: {e}");
+    } else {
+        log::info!("Successfully updated CCADB roots");
+    }
+}
+
+/// Update CCADB roots at each of the provided keys. Roots are never removed
+/// once added.
```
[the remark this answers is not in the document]
lukevalenta: If we really needed to remove some root, we could manually overwrite the roots file in Workers KV, but this would only work if the CCADB list stopped serving the root. Alternatively, we could update the cron job to selectively remove the cert.

bwesterb (approved):
Just some minor comments

PR #83, opened 2025-08-01
Refactor and deduplicate proof logic

lukevalenta:
- Remove duplicate upload_issuers implementations.
- Make ObjectBackend methods parameters more flexible for convenience.
- Rename ProofPreparer -> TlogTileRecorder as this is now used for more than just proofs. (We probably want to move this and PreloadedTlogTileReader to tlog_tiles, but leaving that out of this PR to reduce the diff.)
- Rename SimpleTlogTileReader -> PreloadedTlogTileReader as more descriptive name
- Rename RecordProof -> InclusionProof for consistency with RFC9162 terminology
- Rename TreeProof -> ConsistencyProof for consistency with RFC9162 terminology
- Add `tile_reader_for_indexes` function to deduplicate logic across prove_inclusion, prove_consistency, and read_edge_tiles.
- Use TileHashReader::read_hashes to fetch edge tiles, removing the need for the `read_and_verify_tiles` function and aligning the code more closely to sunlight: https://github.com/FiloSottile/sunlight/blob/409a4d4913368c7e28d012118e567eb1f8c90900/internal/ctlog/ctlog.go#L243.
- Move ProveInclusionQuery/Response structs to mtc frontend worker. Now that prove-inclusion requests are handled without contacting the sequencer, these application-specific endpoints should be defined with the application.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -898,7 +898,6 @@ dependencies = [
  "serde",
  "serde-wasm-bindgen",
  "serde_bytes",
- "serde_with",
  "sha2",
  "signed_note",
  "static_ct_api",
--- a/crates/ct_worker/src/frontend_worker.rs
+++ b/crates/ct_worker/src/frontend_worker.rs
@@ -3,22 +3,16 @@
 
 //! Entrypoint for the static CT submission APIs.
 
-use std::sync::LazyLock;
-
 use crate::{load_roots, load_signing_key, SequenceMetadata, CONFIG};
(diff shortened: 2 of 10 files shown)


lukevalenta:
@rozbb a big one for you!

lukevalenta:
> This would've been easier to review if you had a separate commit for where you move code around; and the one where you renamed things.

Fair enough! Definitely got into the trap of doing too many things at once in this PR. Thanks for the review!

Issue #84, opened 2025-08-02
Tiles missing from R2 storage


lukevalenta:
The raio2025h2 log shard is missing tiles in R2 storage. See https://issues.chromium.org/issues/434895698#comment8 for details.

PR #85, opened 2025-08-02
Double-check that R2 object uploads are successful

lukevalenta:
As a measure to prevent #84, immediately confirm that we can fetch the correct objects from R2 after they are uploaded.

I was having trouble getting the sentry integration working, so in the meantime update the Cleaner to just check that each full level-0 tile exists. This will stop progress indefinitely to allow us to detect if more tiles go missing.

### DIFF
--- a/crates/generic_log_worker/src/lib.rs
+++ b/crates/generic_log_worker/src/lib.rs
@@ -538,15 +538,38 @@ impl ObjectBackend for ObjectBucket {
             .as_ref()
             .inspect(|&m| m.upload_size_bytes.observe(value.len().as_f64()));
         self.bucket
-            .put(key.as_ref(), value)
+            .put(key.as_ref(), value.clone())
             .http_metadata(metadata)
             .execute()
             .await
             .inspect_err(|_| {
                 self.metrics
                     .as_ref()
-                    .inspect(|&m| m.errors.with_label_values(&["put"]).inc());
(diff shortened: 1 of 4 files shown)


crates/generic_log_worker/src/lib.rs
```diff
@@ -547,6 +547,23 @@ impl ObjectBackend for ObjectBucket {
...
+                    "tile persisted with wrong contents: {}",
+                    key.as_ref()
+                )));
+            }
+        } else {
+            return Err(Error::RustError(format!(
+                "tile not persisted: {}",
+                key.as_ref()
+            )));
+        }
```
bwesterb, line 565: We're not checking whether it's correct?

bwesterb: (Just checking it exists is useful for the bug obviously.)

bwesterb:
lgtm

crates/generic_log_worker/src/lib.rs
lukevalenta: We check the contents just above too: https://github.com/cloudflare/azul/pull/85/files/fc116904e7e984997812f14a741038ee235e7345#diff-59a0da9b5d3463610f40f1285b1a51cb2fb2da74ae6799121b2b68899a0efa17R554

lukevalenta:
I made one small tweak to report the R2 errors under the `object_errors_total` metric to allow us to distinguish these R2 errors from other types of 'fatal' sequencing errors.

PR #88, opened 2025-08-04
Fix handling of fatal sequencing error, closes #84

### LINKED ISSUE #84: Tiles missing from R2 storage

lukevalenta:
This bug was introduced in f10355f when trying to reload the sequencer state right away when a fatal sequencing error occurs. If the reloading failed, the old state would not be cleared, as all the calling function does is log the error message.

In this fix, we force the DO to reload the state the next time the alarm or fetch handlers run.

Other (in a separate commit):
- Remove unused dependencies (cargo machete)
- Rename UNHASHED_DATA_TILE_LEVEL -> AUX
- Remove a few unnecessary clones and collects


crates/generic_log_worker/src/sequencer_do.rs
```diff
@@ -165,8 +164,10 @@ impl<L: LogEntry> GenericSequencer<L> {
             &self.metrics,
         )
         .await
+        .is_err()
         {
-            warn!("{name}: Fatal sequencing error, log will be reloaded: {e}",);
+            // Re-initialize the log to get back into a good state.
+            *self.initialized.borrow_mut() = false;
```
bwesterb, line 170: We probably still want to log the error.

PR #91, opened 2025-08-07
Rename `Checkpoint` to `CheckpointText`

rozbb:
Pretty small change. Checkpoints as defined in in [the spec](https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md) are signed notes, which themselves MUST have at least one signature. Thus, the thing we call a checkpoint in the code is actually just the text portion of the checkpoint.

I struggled for a while trying to make a `Checkpoint` type that parses a `CheckpointText` followed by signatures in the signed-note format. But I realized in my second attempt that no such data type is necessary. If you want to produce a checkpoint, you use one of the signing functions, which just returns a `Vec<u8>`. And if you want to parse a checkpoint, you call `open_checkpoint` on a bytestring. And if you _really_ want to open it without verifying it, you can always call `Note::from_reader` and then call `Checkpoint::from_bytes` on `note.text`.

Lmk if you agree @lukevalenta

### DIFF
--- a/crates/generic_log_worker/src/log_ops.rs
+++ b/crates/generic_log_worker/src/log_ops.rs
@@ -1261,7 +1261,7 @@ mod tests {
         PrecertData, StaticCTCheckpointSigner, StaticCTLogEntry, StaticCTPendingLogEntry,
     };
     use std::time::Duration;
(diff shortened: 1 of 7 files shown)


lukevalenta (approved):
The renaming makes sense to me. Thanks!

PR #93, opened 2025-08-25
Ensure that the inferred root is uploaded to /issuers

lukevalenta:
As reported by Andrew Ayer in
https://groups.google.com/a/chromium.org/g/ct-policy/c/dVz6tMLWsRY/m/m1_lh0IzGwAJ, the previous bugfix was incomplete as inferred roots for chains were not getting uploaded to /issuers.

### DIFF
--- a/crates/ct_worker/src/frontend_worker.rs
+++ b/crates/ct_worker/src/frontend_worker.rs
@@ -17,6 +17,7 @@ use static_ct_api::{AddChainRequest, GetRootsResponse, StaticCTLogEntry};
 use tlog_tiles::{LogEntry, PendingLogEntry, PendingLogEntryBlob};
 #[allow(clippy::wildcard_imports)]
 use worker::*;
+use x509_cert::der::Encode;
 
 // The Maximum Merge Delay (MMD) of a log indicates the maximum period of time
 // between when a SCT is issued and the corresponding entry is sequenced in the
@@ -174,6 +175,7 @@ async fn main(req: Request, env: Env, _ctx: Context) -> Result<Response> {
         })
 }
 
+#[allow(clippy::too_many_lines)]
 async fn add_chain_or_pre_chain(
     mut req: Request,
     env: &Env,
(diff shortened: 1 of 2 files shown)


crates/ct_worker/src/frontend_worker.rs
```diff
@@ -232,15 +235,25 @@ async fn add_chain_or_pre_chain(
...
+    // Make sure the root is persisted as well.
+    let root_hash = pending_entry
+        .chain_fingerprints
+        .last()
+        .ok_or("chain fingerprints is empty")?;
+    let Some(root_cert) = roots.by_fingerprint(root_hash) else {
+        return Err(format!("{name}: root hash for chain is not in accepted roots").into());
+    };
+    let root_bytes = root_cert.to_der().map_err(|e| e.to_string())?;
+    if issuers.iter().any(|issuer| *issuer == root_bytes) {
```
bwesterb, line 252: Isn't this the opposite? If none match you want to add?

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -232,15 +235,25 @@ async fn add_chain_or_pre_chain(
...
+        issuers.push(root_bytes.as_slice());
+    }
+
+    generic_log_worker::upload_issuers(&public_bucket, &issuers, name).await?;
```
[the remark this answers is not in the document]
lukevalenta: `upload_issuers` first does a KV fetch to check if the issuer is already there before writing, so there is caching at the KV level.

crates/ct_worker/src/frontend_worker.rs
lukevalenta: Yes, good catch!

lukevalenta:
> This doesn't yet ensure that all roots served by get-roots are present in `/issuers`. Patch coming soon for that.

I added a temporary endpoint to the CT frontend worker to upload roots (as some of them might be missing), applied the fix, and then removed the endpoint. As new entries will now have all issuers correctly uploaded, there's no need to a version of this change around.
```
.get_async("/logs/:log/tmp-upload-roots", |_req, ctx| async move {
    let name = ctx.data;
    let roots = x509_util::certs_to_bytes(&load_roots(&ctx.env, name).await?.certs)
        .map_err(|e| e.to_string())?;
    let object = ObjectBucket::new(load_public_bucket(&ctx.env, name)?);
    generic_log_worker::upload_issuers(
        &object,
        &roots.iter().map(Vec::as_slice).collect::<Vec<&[u8]>>(),
        name.as_ref(),
    )
    .await?;
    Response::ok("successfully uploaded roots")
})
```

PR #96, opened 2025-08-26
Subtree proofs, landmarks, and signatureless certificates

lukevalenta:
This PR adds support for subtree proofs, landmarks, and signatureless certificates. I recommend going commit by commit as some of them do refactoring.

Some features that I haven't added yet, but will file issues for in case we want to add them in the future:
* Make spki_der optional in /get-certificate requests, and fetch the unhashed SPKI from the bootstrap entry to construct the signatureless cert. This would be useful for testing, but probably not necessary for production.
* Add a 'landmark bundle' endpoint to bundle up landmark subtree inclusion proofs to some signed checkpoint. If we do this we _could_ also do the batched consistency proofs (@rozbb has also thought about this, at least for consistency proofs: https://github.com/C2SP/C2SP/pull/18).


crates/tlog_tiles/src/tlog.rs
```diff
@@ -398,34 +415,48 @@ pub fn prove_subtree_inclusion<R: HashReader>(
...
+/// Returns the indexes required for the proof that the tree of size `n`
+/// contains the record with index `leaf_index`.
+///
+/// # Errors
+///
+/// Returns an error if the `[lo, hi)` is not a valid subtree, or if
+/// `leaf_index` is not in that subtree.
+pub fn inclusion_proof_indexes(n: u64, leaf_index: u64) -> Result<Vec<u64>, TlogError> {
+    if leaf_index >= n {
+        return Err(TlogError::InvalidInput("leaf_index < n".into()));
```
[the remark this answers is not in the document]
lukevalenta: Yeah, the printed error is "condition not met: <condition>". I should probably rename the error from `InvalidInput` to `ConditionNotMet` or something so it's more clear.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -435,23 +387,48 @@ fn subtree_hash_internal(lo: u64, hi: u64, hashes: &[Hash]) -> (Hash, Vec<Hash>)
 /// Panics if `read_hashes` returns a slice of hashes that is not the same
 /// length as the requested indexes, or if there are internal math errors.
 pub fn inclusion_proof<R: HashReader>(n: u64, leaf_index: u64, r: &R) -> Result<Proof, TlogError> {
-    if leaf_index >= n {
-        return Err(TlogError::InvalidInput("leaf_index < n".into()));
+    subtree_inclusion_proof(&Subtree::new(0, n)?, leaf_index, r)
+}
+
+/// Returns the subproof that the subtree `n` contains the record with index
```
lukevalenta, line 393: I think this should be 'proof' not 'subproof' here in the comment.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -435,23 +387,48 @@ fn subtree_hash_internal(lo: u64, hi: u64, hashes: &[Hash]) -> (Hash, Vec<Hash>)
...
+/// # Panics
+///
+/// Panics if `read_hashes` returns a slice of hashes that is not the same
+/// length as the requested indexes, or if there are internal math errors.
+pub fn subtree_inclusion_proof<R: HashReader>(
+    n: &Subtree,
+    leaf_index: u64,
+    r: &R,
+) -> Result<Proof, TlogError> {
+    if !n.contains(leaf_index) {
```
lukevalenta, line 409: todo: we should be able to remove the check here, since it'll be checked in `walk_subproof` (via `subproof_indexes`).

crates/tlog_tiles/src/tlog.rs
```diff
@@ -618,25 +543,48 @@ pub fn verify_subtree_inclusion_proof(
...
+///
+/// Panics if `read_hashes` returns a slice of hashes that is not the same
+/// length as the requested indexes, or if there are internal math errors.
+pub fn subtree_consistency_proof<R: HashReader>(
+    tree_size: u64,
+    m: &Subtree,
+    r: &R,
+) -> Result<Proof, TlogError> {
+    let n = Subtree::new(0, tree_size)?;
+    if !n.contains_subtree(m) {
```
lukevalenta, line 569: Should be able to remove this check as well.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -910,6 +794,208 @@ impl Subtree {
...
+        } else {
+            let (mut sibling, mut recursive_proof) = if right.contains_subtree(m) {
+                // Fully include the left, recurse on the right.
+                (f(&left), right.walk_subproof(m, known, f, strategy)?)
+            } else {
+                // `m` is split across children. Fully include the left, recurse
+                // on right child of `m` with `known` set to false as the right
+                // child of `m` was not one of the inputs to the algorithm.
+                assert!(m.lo == self.lo, "bad math in subproof walk");
+                let m_right = m.children().1;
```
bwesterb, line 991: It's not quite obvious this is correct, as it assumes that lo(self.children().1) == lo(m.children().1). In general that's not the case, but here it is, as lo(m)=lo(self) and hi(m) > hi(self.children().0).

crates/tlog_tiles/src/tlog.rs
```diff
@@ -910,6 +794,208 @@ impl Subtree {
...
+    /// In order to calculate either the storage hash indexes needed for the
+    /// subproof or the subproof hashes, this function is generic over:
+    /// - `T`: The type of item to be collected (`u64` or `Hash`).
+    /// - `F`: The type of the closure that performs the action.
+    ///
+    /// When computing indexes, `strategy` indicates how to combine items.
+    /// Indexes are kept in increasing order as that may make retrieval from
+    /// backend storage more efficient, but for hashes the sibling node's hash
+    /// is always appended to the recursive proof according to the spec.
+    fn walk_subproof<T, F>(
```
bwesterb, line 952: It's a bit silly, but are we ever in the case that we want to prove consistency between the empty tree and a non-empty tree?

bwesterb: If this codepath is also used for static-ct, then we do need this. I didn't check whether the order of the consistency proof is that same as the rfc6962.

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -510,53 +510,128 @@ pub enum ProofError {
...
+    let mut indexes = tlog_tiles::subtree_inclusion_proof_indexes(n, leaf_index)?;
+
+    // Fetch the leaf and subtree hashes to verify the proof. These aren't
+    // needed to construct the proof, but fetch them so they're available in
+    // `tile_reader`.
+    indexes.append(&mut tlog_tiles::subtree_hash_indexes(n));
+    let leaf_shx = tlog_tiles::stored_hash_index(0, leaf_index);
+    indexes.push(leaf_shx);
+    indexes.sort_unstable();
+    indexes.dedup();
```
bwesterb, line 565: I don't think there is every duplicates in this case.

crates/generic_log_worker/src/sequencer_do.rs
```diff
@@ -287,3 +288,27 @@ impl<L: LogEntry> GenericSequencer<L> {
...
     }
 }
+
+/// A callback function that gets called each time the sequencer updates the
+/// checkpoint. Currently, this is used only to update the landmark checkpoint
+/// sequence for MTC, but could be extended in the future to collect
+/// cosignatures or perform other application-specific actions.
+pub type CheckpointCallbacker = Box<
+    dyn Fn(
+            u64,
```
bwesterb, line 298: nit: document parameters

crates/mtc_api/src/landmark.rs
```diff
@@ -0,0 +1,241 @@
...
+    last_landmark: usize,
+    landmarks: VecDeque<u64>,
+}
+
+/// The location in object storage for the landmark bundle.
+pub static LANDMARK_KEY: &str = "landmark";
+
+impl LandmarkSequence {
+    /// Create a new landmark sequence with the given `max_landmarks` and an
+    /// initial landmark with id 0 and tree size 0.
```
bwesterb, line 17: Initial tree has size 1, no? There is always the `null` entry.

crates/mtc_api/src/landmark.rs
```diff
@@ -0,0 +1,241 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the list of landmarks is not sorted.
+    pub fn subtree_for_index(&self, leaf_index: u64) -> Option<(usize, Subtree)> {
+        let lo;
+        let hi;
+        let landmark_id;
+        match self.landmarks.binary_search(&leaf_index) {
+            Ok(lo_index) => {
```
bwesterb, line 73: It's a bit annoying we have a special case here. Can we get rid of it?

crates/mtc_api/src/landmark.rs
```diff
@@ -0,0 +1,241 @@
+use anyhow::{anyhow, bail};
+use std::{collections::VecDeque, fmt::Write};
+use tlog_tiles::Subtree;
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct LandmarkSequence {
+    max_landmarks: usize,
+    last_landmark: usize,
+    landmarks: VecDeque<u64>,
+}
```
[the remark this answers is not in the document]
bwesterb: Ah, nvm, didn't read code below yet.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -816,6 +824,32 @@ impl Subtree {
...
+        if hi - lo == 1 {
+            return Ok((Self { lo, hi }, None));
+        }
+        let last = hi - 1;
+        // Find where `lo` and `last`'s tree paths diverge. The two subtrees
+        // will be on either side of the split.
+        // SAFETY: `lo ^ last` is guaranteed to be non-zero, so `ilog2` won't panic.
+        let split = (lo ^ last).ilog2();
+        let mask = (1 << split) - 1;
+        let mid = last & !mask;
```
bwesterb, line 845: `!` is bitwise negation in Rust?

bwesterb (changes requested):
See previous comments on PR.

bwesterb (approved):
Feel free to merge already if that makes future PRs smaller.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -501,73 +501,75 @@ fn inclusion_proof_recursion(
...
+        // a. If sn is 0, then stop the iteration and fail the proof verification.
+        if s_n == 0 {
+            return Err(TlogError::InvalidProof);
+        }
+        // b. If LSB(fn) is set, or if fn is equal to sn, then:
+        if lsb_set(f_n) || f_n == s_n {
+            // i. Set r to HASH(0x01 || p || r).
+            r = node_hash(*p, r);
+            // ii. If LSB(fn) is not set, then right-shift both fn and sn equally until either LSB(fn) is set or fn is 0.
+            while !lsb_set(f_n) || f_n == 0 {
```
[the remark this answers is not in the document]
lukevalenta: This clarifies the text in the spec: https://github.com/davidben/merkle-tree-certs/pull/139.

crates/tlog_tiles/src/tlog.rs
```diff
@@ -692,80 +694,188 @@ fn consistency_proof_recursion(
...
+    pub fn new(lo: u64, hi: u64) -> Result<Self, TlogError> {
+        if lo >= hi {
+            return Err(TlogError::InvalidInput("`lo < hi`".into()));
+        }
+        // `s` is the smallest power of 2 that is greater than or equal
+        // to `lo - hi`.
+        let s = {
+            let n = hi - lo;
+            let l = n.ilog2();
+            // If n is not already a power of two, round up.
```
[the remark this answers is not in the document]
lukevalenta: Yep, nice simplification! Just one tweak: in python it would be `1 << (n-1).bit_width()` since `n.bit_width()` is `n.ilog2() + 1` (e.g., `1.bit_width()  == 1`, but `1.ilog2() == 0`)

crates/tlog_tiles/src/tlog.rs
```diff
@@ -816,6 +824,32 @@ impl Subtree {
...
+            return Err(TlogError::InvalidInput("`lo < hi`".into()));
+        }
+        if hi - lo == 1 {
+            return Ok((Self { lo, hi }, None));
+        }
+        let last = hi - 1;
+        // Find where `lo` and `last`'s tree paths diverge. The two subtrees
+        // will be on either side of the split.
+        // SAFETY: `lo ^ last` is guaranteed to be non-zero, so `ilog2` won't panic.
+        let split = (lo ^ last).ilog2();
```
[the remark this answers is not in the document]
lukevalenta: ilog2() should be the same as bit_length()-1

crates/tlog_tiles/src/tlog.rs
lukevalenta: Yep!

crates/mtc_api/src/landmark.rs
lukevalenta: This comes from the spec: "Each landmark specifies a tree size. The first landmark, numbered zero, is always a tree size of zero."

But since landmark zero is inactive, I don't think it matters if it covers the `null` entry or not.

crates/tlog_tiles/src/tlog.rs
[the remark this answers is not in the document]
bwesterb: Ah, yes, you're right. (Unless of course for the zero case which is excluded.)

crates/mtc_api/src/landmark.rs
bwesterb: My reasoning for including it, is that a zero length subtree is a bit annoying to deal with and we're avoiding it elsewhere.

bwesterb: Ah, ok, never mind then!

crates/mtc_api/src/landmark.rs
lukevalenta: Yes, good point! Just pushed a big simplification

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -510,53 +510,128 @@ pub enum ProofError {
...
+    let n = &Subtree::new(start, end)?;
+    let mut indexes = tlog_tiles::subtree_inclusion_proof_indexes(n, leaf_index)?;
+
+    // Fetch the leaf and subtree hashes to verify the proof. These aren't
+    // needed to construct the proof, but fetch them so they're available in
+    // `tile_reader`.
+    indexes.append(&mut tlog_tiles::subtree_hash_indexes(n));
+    let leaf_shx = tlog_tiles::stored_hash_index(0, leaf_index);
+    indexes.push(leaf_shx);
+    indexes.sort_unstable();
```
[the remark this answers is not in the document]
lukevalenta: See https://github.com/cloudflare/azul/pull/96#discussion_r2307902700 (let's discuss in the other thread)

crates/mtc_api/src/lib.rs
```diff
@@ -44,6 +48,65 @@ use x509_util::CertPool;
...
+    ObjectIdentifier::new_unwrap("1.3.6.1.4.1.44363.47.0");
+
+// MTCSignature from <https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-05.html#name-certificate-format>.
+struct MtcSignature {
+    cosigner_id: TrustAnchorID,
+    signature: Vec<u8>,
+}
+
+impl MtcSignature {
+    fn serialize(&self) -> Vec<u8> {
```
rozbb, line 63: `to_bytes`

crates/mtc_api/src/lib.rs
```diff
@@ -44,6 +48,65 @@ use x509_util::CertPool;
...
+// MTCProof from <https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-05.html#name-certificate-format>.
+struct MtcProof {
+    start: u64,
+    end: u64,
+    inclusion_proof: Proof,
+    signatures: Vec<MtcSignature>,
+}
+
+impl MtcProof {
+    fn serialize(&self) -> Vec<u8> {
```
rozbb, line 82: `to_bytes`

crates/generic_log_worker/src/sequencer_do.rs
```diff
@@ -287,3 +288,32 @@ impl<L: LogEntry> GenericSequencer<L> {
         Response::ok(buffer)
     }
 }
+
+/// A callback function that gets called each time the sequencer updates the
+/// checkpoint. Currently, this is used only to update the landmark checkpoint
+/// sequence for MTC, but could be extended in the future to collect
+/// cosignatures or perform other application-specific actions.
```
rozbb, line 295: It's kinda surprising that this is an `Fn` that returns nothing. You'd think this function cannot do anything. Maybe add a note in the docs here that it's expected to capture stuff with interior mutability, eg buckets, so that it can modify the environment.

crates/mtc_worker/src/sequencer_do.rs
```diff
@@ -49,3 +52,60 @@ impl DurableObject for Sequencer {
...
+
+/// Return a callback function that gets passed into the generic sequencer and
+/// called each time a new checkpoint is created. For MTC, this is used to
+/// periodically update the landmark checkpoint sequence.
+fn checkpoint_callback(env: &Env, name: &str) -> CheckpointCallbacker {
+    let params = &CONFIG.logs[name];
+    let bucket = load_public_bucket(env, name).unwrap();
+    Box::new(
+        move |tree_size: u64, old_time: UnixTimestamp, new_time: UnixTimestamp| {
+            Box::pin({
```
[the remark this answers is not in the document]
rozbb: That's my understanding too. You cannot leak unless there's an `unsafe` somewhere, right?

crates/tlog_tiles/src/tlog.rs
```diff
@@ -380,387 +402,642 @@ pub fn prove_inclusion<R: HashReader>(t: u64, n: u64, r: &R) -> Result<Inclusion
...
-    th: Hash,
-    n: u64,
-    h: Hash,
+/// Will panic if there are internal math errors.
+pub fn verify_inclusion_proof(
+    proof: &Proof,
+    tree_size: u64,
+    root_hash: Hash,
+    leaf_index: u64,
+    leaf_hash: Hash,
```
rozbb, line 475: Ty for the rename

crates/generic_log_worker/src/sequencer_do.rs
lukevalenta: Ah yeah after the worker 0.6.0 release there's a ton of interior mutability in workers-rs. I'll add a note!

crates/mtc_api/src/landmark.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    /// Serialize according to
+    /// <https://www.ietf.org/archive/id/draft-davidben-tls-merkle-tree-certs-06.html#section-6.3.1>.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if writing to the buffer fails.
+    pub fn serialize(&self) -> Result<Vec<u8>, anyhow::Error> {
```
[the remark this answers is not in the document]
lukevalenta: Sounds good! I never have a great sense of whether to use `serialize`, `unmarshal`, or just `to_bytes`, but that seems like a good rule.

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -1383,7 +1524,7 @@ mod tests {
                 &log.object,
             ))
             .unwrap();
-            tlog_tiles::check_consistency(
+            tlog_tiles::verify_consistency_proof(
```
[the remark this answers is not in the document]
lukevalenta: I'm not sure I follow. What's the trivial consistency proof, and what's the assert?

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -493,75 +492,150 @@ impl SequenceState {
...
-    let indexes = tlog_tiles::inclusion_proof_indexes(0, tree_size, leaf_index, Vec::new());
-    let tile_reader = tile_reader_for_indexes(tree_size, &indexes, object)
-        .await
-        .map_err(|e| e.to_string())?;
+    let n = &Subtree::new(start, end)?;
+    let mut indexes = tlog_tiles::subtree_inclusion_proof_indexes(n, leaf_index)?;
+
+    // Fetch the leaf and subtree hashes to verify the proof. These aren't
+    // needed to construct the proof, but fetch them so they're available in
+    // `tile_reader`.
```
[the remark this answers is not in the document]
lukevalenta: Yeah, the `tlog_tiles::subtree_hash` function expects the tiles to already be available from the reader, and errors if it's not. We can't just fetch the tiles within that function since it's sync and we need async to fetch from R2.

Regardless, I'll chop this code to do proof validation from the `prove_inclusion` function, since it's not really necessary.

crates/generic_log_worker/src/log_ops.rs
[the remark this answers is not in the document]
rozbb: There was an old assert that the old tree size must be between 0 and the new tree size minus 1. That assert was deleted. So we should make sure that this works when the old tree size equals the new tree size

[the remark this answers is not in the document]
lukevalenta: Ah right, so all of the checks now funnel to https://github.com/cloudflare/azul/pull/96/files#diff-f7c4a6f74d762dbdcb57c48fac43474ad29c9a5454d894809c866ac5164d4b5dR961 (which replaces the assertions with a thrown error), and I believe the test at https://github.com/cloudflare/azul/pull/96/files#diff-f7c4a6f74d762dbdcb57c48fac43474ad29c9a5454d894809c866ac5164d4b5dR1228 covers the trivial consistency proof cases where the trees are of equal size. (Since `j` ranges up to `i`, and we're checking the trees of size `i+1` and `j+1`).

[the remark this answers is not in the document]
rozbb: Ah ok great

cjpatton (commented):
Working my way through, just a few minor things so far.

crates/mtc_api/src/landmark.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+    /// Add a new landmark with the given tree size, removing a landmark if the
+    /// maximum size would be exceeded. Returns true if the new landmark is
+    /// added, or false otherwise.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if the tree size is smaller than the last landmark
+    /// tree size.
+    pub fn add(&mut self, tree_size: u64) -> Result<bool, anyhow::Error> {
```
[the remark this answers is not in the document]
lukevalenta: Oh yep that was just some laziness. No need for anyhow here!

crates/tlog_tiles/src/tlog.rs
```diff
@@ -362,12 +367,40 @@ pub fn subtree_indexes(lo: u64, hi: u64, mut need: Vec<u64>) -> Vec<u64> {
...
+pub fn subtree_hash<R: HashReader>(n: &Subtree, r: &R) -> Result<Hash, TlogError> {
+    let indexes = subtree_indexes_internal(n.lo, n.hi, vec![]);
+    let hashes = r.read_hashes(&indexes)?;
+    assert_eq!(
+        hashes.len(),
+        indexes.len(),
+        "bad read_hashes implementation"
+    );
+    let (hash, remaining_hashes) = subtree_hash_internal(n.lo, n.hi, &hashes);
+    assert!(remaining_hashes.is_empty(), "bad math in tree_hash");
```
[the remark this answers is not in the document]
lukevalenta: Oh whoops. This is already fixed in a subsequent commit as all of the `_internal` functions get replaced.

PR #111, opened 2025-09-12
Keep only a single copy of roots per worker script, fixes #110

### LINKED ISSUE #110: CT log gets into degraded state due to memory pressure
From the Raio2026h2a CT log returned almost exclusively 429 errors (https://issues.chromium.org/issues/434895698#comment14), and we saw similar patches of high errors rates (sometimes manifesting as 500 errors instead) from the Raio2025h2b log. From internal metrics it appears that the errors appears due to the workers isolate reaching memory limits, and the wrangler tail logs confirm this during periods of high error rates.

```
POST https://ct.cloudflare.com/logs/raio2025h2b/ct/v1/add-chain - Exception Thrown @ 9/4/2025, 2:31:03 PM
✘ [ERROR]   Error: memory access out of bounds 
```

The errors started August 24, soon after https://github.com/cloudflare/azul/commit/ad6ced9d2b0a7850230777b8299a47120640b330 was deployed, which fixes the issue reported in https://groups.google.com/a/chromium.org/g/ct-policy/c/dVz6tMLWsRY/m/ApNZBFkjGwAJ. We decided not to roll back as that would re-introduce the bug. We've been able to reproduce the issue for the cftest log shards, and confirmed that rolling back does fix the memory issues.

Closer inspection of the commit in question reveals that we keep the 'roots' variable (containing all accepted roots for the log) in scope from https://github.com/cloudflare/azul/commit/ad6ced9d2b0a7850230777b8299a47120640b330#diff-130a857a7156f3c81dd0e04249775b84bf4439aed5846bddaa27337e84ce06c4R195 all the way until the return at https://github.com/cloudflare/azul/commit/ad6ced9d2b0a7850230777b8299a47120640b330#diff-130a857a7156f3c81dd0e04249775b84bf4439aed5846bddaa27337e84ce06c4R311, instead of immediately releasing the variable from scope so that the memory is de-allocated. Specifically, the variable is kept in memory across the asynchronous call to the sequencer, which typically takes 1-2 seconds. Across many requests, memory is quickly exhausted, and DOs also seem to be impacted.

lukevalenta:
In 36bb0fc, we switched from loading roots once per Worker script (with
LazyLock) to loading roots dynamically for each fetch request, since we
need an async operation to fetch roots from KV, and LazyLock doesn't
support async initialization. This means each add-chain request has to
keep a separate copy of the roots in the Worker memory. Combined with
ad6ced9 which caused the dynamically loaded roots to be kept in memory
across the asynchronous call to the sequencer, this resulted in the
memory issues observed in #110.

Use the runtime-agnostic tokio::sync::LazyCell to restore the previous
behavior of loading roots only once for each Worker script.

This is deployed for the cftest logs, and we are no longer seeing high rates of 500/429 errors.

This PR contains several other commits that can be reviewed independently, including two bug fixes in chain validation (none critical).

### DIFF
--- a/crates/ct_worker/src/frontend_worker.rs
+++ b/crates/ct_worker/src/frontend_worker.rs
@@ -204,7 +204,7 @@ async fn add_chain_or_pre_chain(
 
     // Temporal interval dates prior to the Unix epoch are treated as the Unix epoch.
     let roots = load_roots(env, name).await?;
-    let pending_entry = match static_ct_api::partially_validate_chain(
+    let (pending_entry, found_root_idx) = match static_ct_api::partially_validate_chain(
         &req.chain,
         roots,
         Some(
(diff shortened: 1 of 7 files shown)


bwesterb (approved):

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -244,22 +244,24 @@ async fn add_chain_or_pre_chain(
 
     // Entry is not cached, so we need to sequence it.
 
-    // First persist issuers.
-    let public_bucket = ObjectBucket::new(load_public_bucket(env, name)?);
-    let mut issuers = req.chain[1..]
-        .iter()
-        .map(Vec::as_slice)
-        .collect::<Vec<&[u8]>>();
+    // First persist issuers. Use a block so memory is deallocated sooner.
```
cjpatton, line 247: One downside of implicit alloc/dealloc :)

Issue #109, opened 2025-09-11
Validate certificates in MTC worker


rozbb:
There should be a basic validity check of bootstrap certs before sequencing them [here](https://github.com/cloudflare/azul/blob/1e563ca7070ad89b7f4b0929fd01d036bb2f2e28/crates/mtc_api/src/lib.rs#L503). Can be similar to the code in [static CT](https://github.com/cloudflare/azul/blob/1e563ca7070ad89b7f4b0929fd01d036bb2f2e28/crates/ct_worker/src/frontend_worker.rs#L207).

lukevalenta:
Resolved in https://github.com/cloudflare/azul/pull/113.

PR #113, opened 2025-09-16
MTC worker: validate bootstrap certs

rozbb:
Currently, the `validate_chain` function in the Merkle Tree Certificate worker does not verify signatures on the provided bootstrap certificate chain (#109). This is not a security issue, but it does mean the log can get spammed easier because it will put obviously bad certificates on the log.

This PR:
1. Moves the common validation logic from `static_ct_api` to `x509_util`
2. Rewrites domain-specific validation logic as a hook that `static_ct_api` and `mtc_worker` give to the generic validator

One question I had: what precisely is [`validate_correspondence()`](https://github.com/cloudflare/azul/blob/1e563ca7070ad89b7f4b0929fd01d036bb2f2e28/crates/mtc_api/src/lib.rs#L503) exposed for? I didn't implement validation there because it had a slightly different type signature, and it didn't look like it was used anywhere. I can def add it if it's needed though.


crates/generic_log_worker/src/lib.rs
```diff
@@ -554,6 +554,34 @@ impl ObjectBackend for ObjectBucket {
                 .observe(millis_diff_as_secs(start, now_millis()));
         });
 
+        // SAFETY: immediately fetch the uploaded object to make sure that it was persisted.
```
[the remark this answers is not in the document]
rozbb: Deleted

crates/mtc_api/src/lib.rs
```diff
@@ -498,119 +500,136 @@ pub fn tbs_cert_to_log_entry(
...
         }
-        _ => return Err(MtcError::Dynamic("mismatched extensions".into())),
-    }
+        Ok(())
+    };
 
-    Ok(())
+    // Run the validation logic with the above validation hook
+    validate_chain_lax(raw_chain, roots, None, None, validator_hook).map_err(|e| match e {
+        x509_util::HookOrValidationError::Valiadation(ve) => ve.into(),
```
lukevalenta, line 630:
```suggestion
        x509_util::HookOrValidationError::Validation(ve) => ve.into(),
```

crates/static_ct_api/src/rfc6962.rs
```diff
@@ -79,12 +78,11 @@ pub struct GetRootsResponse {
 
 /// Validates a certificate chain according to
 /// [RFC6962](https://datatracker.ietf.org/doc/html/rfc6962) and returns a
-/// pending log entry.
+/// pending log entry and the inferred root, if a root had to be inferred.
```
lukevalenta, line 81:
```suggestion
/// pending log entry and the index in the cert pool of the inferred root, if a root had to be inferred.
```

crates/static_ct_api/src/rfc6962.rs
```diff
@@ -93,154 +91,105 @@ pub fn partially_validate_chain(
...
-    // CT is intended to observe certifications rather than police them.
-
-    // Keep an owned copy of the parsed certs, but use a vector of references so
-    // we can avoid needing to clone the inferred root cert when adding it to
-    // the chain.
-    let chain_certs_owned: Vec<Certificate> = raw_chain[1..]
-        .iter()
-        .map(|v| parse_certificate(v))
-        .collect::<Result<_, _>>()?;
-    let mut chain_certs = chain_certs_owned.iter().collect::<Vec<_>>();
```
lukevalenta, line 144: If we bring this change back in (from https://github.com/cloudflare/azul/commit/72843c9acf4f262b1ca6a38f24c6578e64cb84f0), we should be able to avoid a clone and just pass the intermediates (or chain_certs) around by reference.

crates/static_ct_api/src/rfc6962.rs
```diff
@@ -93,154 +91,105 @@ pub fn partially_validate_chain(
...
-            if chain_certs.len() < 2 {
-                return Err(StaticCTError::MissingPrecertSigningCertificateIssuer);
+        // Construct a pending log entry.
+        let precert_opt: Option<PrecertData>;
+        let certificate: Vec<u8>;
+        if is_leaf_precert {
+            let mut pre_issuer: Option<&TbsCertificate> = None;
+            let issuer_key_hash: [u8; 32];
+            if has_precert_signing_cert {
+                pre_issuer = Some(&intermediates[0].tbs_certificate);
```
lukevalenta, line 140: Could you rename this back to `chain_certs` to reduce the PR diff?

lukevalenta (changes requested):

crates/static_ct_api/src/rfc6962.rs
rozbb: I do do that now [here](https://github.com/cloudflare/azul/pull/113/files#diff-512837efef7e7dce3992999e4e7562757d436db5a3511cc285908b7b47455694R221-R226), is that what you meant?

crates/x509_util/src/lib.rs
```diff
@@ -133,3 +134,297 @@ impl CertPool {
...
+/// hook errors.
+pub fn validate_chain_lax<T, E, F>(
+    raw_chain: &[Vec<u8>],
+    roots: &CertPool,
+    not_after_start: Option<UnixTimestamp>,
+    not_after_end: Option<UnixTimestamp>,
+    hook: F,
+) -> Result<T, HookOrValidationError<E>>
+where
+    F: FnOnce(Certificate, Vec<Certificate>, Vec<[u8; 32]>, Option<usize>) -> Result<T, E>,
```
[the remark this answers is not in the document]
rozbb: It's actually less cloning if we pass by reference. Because the hook has to return some values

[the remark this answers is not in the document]
lukevalenta: Ah right, since the hook is returning a PendingLogEntry

crates/static_ct_api/src/rfc6962.rs
```diff
@@ -93,154 +91,105 @@ pub fn partially_validate_chain(
...
+                if intermediates.len() < 2 {
```
lukevalenta, line 141: I believe at this point `intermediates` does not include the inferred root (since we're passing around `inferred_root_idx` instead), but these checks assume that the inferred root is appended.

rozbb: Great catch. I don't know why this didn't error. Maybe worth a regression test in the future

crates/x509_util/src/lib.rs
```diff
@@ -133,3 +134,293 @@ impl CertPool {
...
+        // Check that intermediates have the CA Basic Constraint.
+        // Precertificate signing certificates must also have CA:true.
+        if cert
+            .tbs_certificate
+            .get::<BasicConstraints>()
+            .map_err(ValidationError::from)?
+            .is_none_or(|(_, bc)| !bc.ca)
+        {
+            return Err(ValidationError::IntermediateMissingCaBasicConstraint.into());
+        }
```
rozbb, line 253: Question: the above code block cites CT for this check, but it seems very normal. Should we keep this in the X509 generic check or move it to the CT crate? @lukevalenta

lukevalenta (approved):

PR #126, opened 2025-10-02
Add option to stop chain validation when a trusted path is found

lukevalenta:
This option allows the Merkle Tree CA to validate cross-signed bootstrap chains, where the final cert in the chain is not necessarily trusted. This option is disabled for CT where every cert in the submitted chain needs to be part of the validated path.

Also enforce path length constraints in validate_chain_lax. RFC 6962 doesn't mention invalid path length constraints as a reason to reject a certificate chain, but it is required in RFC 9162. Seems fine to reject these chains in CT. For MTC we do want to enforce this check.

### DIFF
--- a/crates/x509_util/src/lib.rs
+++ b/crates/x509_util/src/lib.rs
@@ -230,6 +230,11 @@ pub struct ValidationOptions {
 ///
 /// Returns a `ValidationError` if the chain fails to validate. Returns an error
 /// of type `E` if the hook errors.
+///
+/// # Panics
+///
+/// This section is here to avoid linter complaints about the `unwrap()` below,
+/// but this function won't panic.
 pub fn validate_chain_lax<T, E, F>(
     raw_chain: &[Vec<u8>],
     roots: &CertPool,
@@ -296,8 +301,8 @@ where
         // Add the intermediate to the validated chain.
         validated_intermediates.push(intermediate_cert);
 
-        // Check this intermediate next.
-        current_cert = &validated_intermediates[i];
+        // Get a reference to the intermediate we just pushed to check next.
+        current_cert = validated_intermediates.last().unwrap();
     }
 
     // If we haven't yet found a path to a trusted root, check if we can find


crates/x509_util/src/lib.rs
```diff
@@ -236,87 +255,145 @@ where
...
-            .find_potential_parents(to_verify)
-            .map_err(ValidationError::from)?
-            .iter()
-            .find(|&&roots_idx| is_link_valid(to_verify, &roots.certs[roots_idx]))
+    // If we haven't yet found a path to a trusted root, check if we can find
+    // one for the last cert in the chain. If we can't, fail chain validation.
+    let path_to_root = if let Some(path) = path_to_root {
+        path
+    } else {
+        let Some(path) = find_path_to_root(current_cert, roots, validated_intermediates.len())?
```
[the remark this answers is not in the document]
lukevalenta: Yeah, agree on the confusing control flow. I tried a few different things, but this ended up looking the simplest.

> If we stopped on first valid cert, then `path_to_root` will be `Some(...)`
Yes. This can only happen if `stop_on_first_trusted_cert` is true _and_ there is at least one intermediate in the chain (since otherwise we skip the loop) _and_ we stop the loop early before we process the last cert in the chain.

> Otherwise, `path_to_root` will already be set according to the last cert in the chain.

If it's not the case that all of the above conditions hold, then `path_to_root` will still be `None` when we reach this line and we still need to look for a path to the trusted root.

cjpatton (approved):

crates/x509_util/src/lib.rs
```diff
@@ -236,87 +255,145 @@ where
...
+/// # Arguments
+///
+/// * `cert` - The cert for which to find a path to a trusted root.
+/// * `roots` - The cert pool of trusted roots.
+/// * `num_intermediates` - The number of intermediate certs preceding the cert
+///   in the chain. This is used for checking the path length basic constraint.
+fn find_path_to_root(
+    cert: &Certificate,
+    roots: &CertPool,
+    num_intermediates: usize,
```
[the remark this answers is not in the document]
lukevalenta: Oh I misread your original comment -- yes, they're the same thing and I'll update the variable name in `check_ca_basic_constraints` 👍

crates/x509_util/src/lib.rs
```diff
@@ -236,87 +255,156 @@ where
...
-            .is_none_or(|(_, bc)| !bc.ca)
-        {
-            return Err(ValidationError::IntermediateMissingCaBasicConstraint.into());
-        }
+
+        // Add the intermediate to the validated chain.
+        validated_intermediates.push(intermediate_cert);
+
+        // Check this intermediate next.
+        current_cert = &validated_intermediates[i];
```
rozbb, line 300: Nit: maybe write `current_cert = validated_intermediates.last().as_ref().unwrap()` to convey that it's indeed the thing you just pushed

crates/x509_util/src/lib.rs
```diff
@@ -236,87 +255,156 @@ where
...
+            PathToRoot::SignedByRoot(root_pool_idx) => {
+                // Append the trusted root to the validated chain.
+                let root = &roots.certs[root_pool_idx];
+                let root_der = root.to_der().map_err(ValidationError::from)?;
+                chain_certs.push(root);
+                chain_fingerprints.push(Sha256::digest(&root_der).into());
+                Some(root_pool_idx)
+            }
+        }
+    };
```
rozbb, line 352: This function is looking pretty clean!

crates/x509_util/src/lib.rs
```diff
@@ -236,87 +255,156 @@ where
...
+    num_intermediates: usize,
+) -> Result<Option<PathToRoot>, ValidationError> {
+    // Check if the cert is itself a trusted root.
+    if roots.includes(cert).map_err(ValidationError::from)? {
+        return Ok(Some(PathToRoot::IsRoot));
+    }
+
+    // Check if the cert is signed by a trusted root.
+    if let Some(&found_idx) = roots
+        .find_potential_parents(cert)
```
rozbb, line 386: Note to self: this is not a linear scan. If it were, the runtime of validation would be c*r where c is the chain length and r is the number of roots. Instead, this is a hashmap lookup

rozbb (approved):

crates/x509_util/src/lib.rs
lukevalenta: Done!

PR #127, opened 2025-10-06
Unify branches in subtree consistency proof

lukevalenta:
See https://github.com/davidben/merkle-tree-certs/pull/149 for details. As an added bonus, this removes a `clone()` of the proof.

This also handles an underflow panic when `n` is zero.


rozbb (approved):
Looks like it follows the steps in the docs!

PR #128, opened 2025-10-06
mtc_worker: Add API for the current landmark bundle

### LINKED ISSUE #104: Add a 'landmark bundle' endpoint to bundle up landmark subtree inclusion proofs to some signed checkpoint
If we do this, we could optionally make it a bulk subtree consistency proof (with deduplicated tree nodes), but gzipping the bundle is probably just as effective.

cjpatton:
Closes #104.

A "landmark bundle" consists of the minimum amount of information needed to validate all MTCs issued before some checkpoint. It has the following components:

1. The checkpoint, including co-signatures

2. The sequence of landmark subtrees determined by the landmark sequence up to the checkpoint

3. A proof of consistency of each subtree with the checkpoint.

Add an API to mtc_worker for fetching the current landmark bundle.

### DIFF
--- a/crates/mtc_api/src/landmark.rs
+++ b/crates/mtc_api/src/landmark.rs
@@ -165,6 +165,49 @@ impl LandmarkSequence {
             landmarks,
         })
     }
+
+    /// Iterate over the sequence of subtrees determined by the landmark sequence.
+    pub fn subtrees(&self) -> LandmarkSubtreesIterator<'_> {
+        LandmarkSubtreesIterator {
+            index: 1,
+            landmarks: &self.landmarks,
+            next_subtree: None,
+        }
+    }
+}
+
+/// An iterator over the subtrees determined by the landmark sequence.
+pub struct LandmarkSubtreesIterator<'a> {
+    index: usize,
+    landmarks: &'a VecDeque<u64>,
+    next_subtree: Option<Subtree>,
+}
+
+impl Iterator for LandmarkSubtreesIterator<'_> {
+    type Item = Subtree;
+
+    fn next(&mut self) -> Option<Subtree> {
+        if self.landmarks.len() < 2 {
+            return None;
+        }
+
+        if let Some(subtree) = self.next_subtree.take() {
+            self.next_subtree = None;
+            return Some(subtree);
+        }
+
+        if self.index == self.landmarks.len() {
+            return None;
+        }
+
+        let subtree;
+        (subtree, self.next_subtree) =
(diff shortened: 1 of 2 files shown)


crates/mtc_worker/src/frontend_worker.rs
```diff
@@ -442,6 +441,93 @@ async fn add_entry(mut req: Request, env: &Env, name: &str) -> Result<Response>
...
+        {
+            Ok(p) => p,
+            Err(ProofError::Tlog(s)) => return Response::error(s.to_string(), 422),
+            Err(ProofError::Other(e)) => return Err(e.to_string().into()),
+        };
+
+        subtrees.push(SubtreeWithConsistencyProof {
+            start: landmark_subtree.lo(),
+            end: landmark_subtree.hi(),
+            consistency_proof: consistency_proof
```
lukevalenta, line 474: nit: Right now this response is sort of half JSON and half binary which feels a little weird. Making the proof an array of base64 hashes like https://datatracker.ietf.org/doc/html/rfc6962#section-4.5 would make manual inspection easier, or maybe we should ditch the JSON and put this all into a binary format.

crates/mtc_api/src/landmark.rs
```diff
@@ -165,6 +165,49 @@ impl LandmarkSequence {
...
+}
+
+/// An iterator over the subtrees determined by the landmark sequence.
+pub struct LandmarkSubtreesIterator<'a> {
+    index: usize,
+    landmarks: &'a VecDeque<u64>,
+    next_subtree: Option<Subtree>,
+}
+
+impl Iterator for LandmarkSubtreesIterator<'_> {
```
lukevalenta, line 186: This is nice.

lukevalenta (approved):
A couple bonus changes made it in, but otherwise looks good!

crates/mtc_worker/ls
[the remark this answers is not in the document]
cjpatton: 😮‍💨

crates/mtc_worker/src/frontend_worker.rs
cjpatton: Ah, I was going for consistency with [`MtcProof`](https://github.com/cloudflare/azul/blob/63ec7b5c3292d5bbc2654e833e98941eff7eb46d/crates/mtc_api/src/lib.rs#L87-L91), but consistency with RFC6962 makes more sense. Done.

lukevalenta (approved):

PR #150, opened 2025-10-30
mtc_worker: Add subtree hashes to landmark bundle

cjpatton:


bwesterb (approved):

bwesterb:
Note that this doesn't match the [documented interface](https://github.com/davidben/merkle-tree-certs/blob/main/draft-davidben-tls-merkle-tree-certs.md#landmarks) to expose landmarks.

bwesterb:
I missed this in the previous round of reviews, but it's not ideal that there is quite a bit of duplication in this API: we need to check that the end of one subtree matches the start of the next; and that the treesize of the checkpoint matches end-1 of the last subtree.

PR #157, opened 2025-11-10
mtc_worker: Define shard3

cjpatton:


bwesterb:
What landmark base ID do we use here?

How long do we keep the old shards around?

bwesterb (approved):

cjpatton:
Rebased.

PR #161, opened 2025-11-13
ct-worker: Fix local testing

### LINKED ISSUE #160: ct-worker Local Deployment Instructions Fail
Following the [Local deployment instructions](https://github.com/cloudflare/azul/tree/7ca0f43b5f66fb9cb920a57621f13e7585b1c95e/crates/ct_worker#local-deployment) in the ct-worker crate results in a "Bad request" response on the curl command. This makes it difficult to locally test changes to azul.

Will investigate.

rozbb:
Previously, following the `openssl` bash line in the "local deployment section" was failing because it uses a hard-coded shard, regardless of the certificate's not-after timestamp.

This PR:
1. Replaces the bash line with a Python script that picks the shard correctly
2. Extends our dev shards into 2027
3. Extends local testing dev shards into 2027
4. Updates the script that creates shards to have better error handling and documentation

Fixes #160

### DIFF
--- a/crates/ct_worker/scripts/add_cert_to_local_dev.py
+++ b/crates/ct_worker/scripts/add_cert_to_local_dev.py
@@ -35,8 +35,9 @@ log_name = f"dev{not_after_year}h{not_after_half}a"
 
 # Make a POST request to localhost add-chain
 payload = {"chain": b64_cert_chain}
+url = f"http://localhost:8787/logs/{log_name}/ct/v1/add-chain"
 request = urllib.request.Request(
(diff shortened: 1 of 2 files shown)


crates/ct_worker/scripts/add_cert_to_local_dev.py
```diff
@@ -0,0 +1,51 @@
...
+import json
+import urllib.request
+import socket
+import ssl
+import sys
+
+HOST = "www.google.com"
+
+# Need >=3.13 to use conn.get_unverified_chain()
+if sys.version_info < (3, 13, 0):
```
cjpatton: cool, TIL about tuple total ordering!

PR #169, opened 2025-11-28
Improve utility scripts

mendess:
- **Make the create log script edit the wrangler.jsonc**
    This way it's a lot easier to create new logs and make sure the
    configuration is updated correctly

- **Add script to create the ccadb_roots kv namespace**
    To serve as both utility and documentation

### DIFF
--- a/crates/ct_worker/scripts/create-root-kv.sh
+++ b/crates/ct_worker/scripts/create-root-kv.sh
@@ -5,15 +5,15 @@ cd "$(dirname "$0")/.." || exit # this script assumes it's runnnig inside the ct
 
 # Helper script to create resources for a log shard.
 
-if [ -z $ENV ] || [ -z $CLOUDFLARE_ACCOUNT_ID ]; then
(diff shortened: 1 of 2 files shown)


crates/ct_worker/scripts/create-root-kv.sh
```diff
@@ -0,0 +1,31 @@
...
+while true; do
+    read -rp "Do you want to proceed with ENV=${ENV}, CLOUDFLARE_ACCOUNT_ID=${CLOUDFLARE_ACCOUNT_ID}? (y/N) " yn
+    case $yn in
+        [yY] ) echo "Proceeding..."; break;;
+        [nN] ) echo "Exiting..."; exit;;
+        * ) echo "Invalid input. Please enter 'y' or 'N'.";;
+    esac
+done
+
+# Create KV namespace if it does not already exist
```
rozbb: The `create-log` file usually does an `if` statement before this. Is `kv namespace create` automatically non-overwrite? If so, it'd be nice to have that in the comment above

cjpatton (approved):

PR #170, opened 2025-12-24
Add WASM-compatible SCT validation with Chrome CT policy

lbaquerofierro:
## Summary

- Add `sct_validator` crate for validating Signed Certificate Timestamps (SCTs) embedded in certificates
- Implement Chrome's Certificate Transparency policy requirements
- Integrate optional SCT validation into the mtc_worker submission flow
- Add cron job to fetch and cache Google's CT log list

### New `sct_validator` crate
A WASM-compatible SCT validation library that implements Chrome's CT policy:
- Certificates ≤ 180 days require 2 SCTs from unique logs
- Certificates > 180 days require 3 SCTs from unique logs
- Always requires SCTs from 2 different log operators
- Handles log states: Qualified, Usable, ReadOnly (rejects Pending/Retired)
- Auto-succeeds if log list is stale (> 70 days old)

### mtc_worker integration
- New `ct_logs_cron.rs`: Fetches Chrome's CT log list and caches in KV
- New config option `enable_sct_validation` (default: false)
- Frontend validates embedded SCTs before accepting certificates when enabled
- Added `ct_logs` KV namespace binding in wrangler.jsonc


crates/mtc_worker/src/ccadb_roots_cron.rs
```diff
@@ -37,6 +37,24 @@ async fn main(_event: ScheduledEvent, env: Env, _ctx: ScheduleContext) {
...
+                crate::ct_logs_cron::CT_LOGS_NAMESPACE
+            );
+            return;
+        }
+    };
+    if let Err(e) = crate::ct_logs_cron::update_ct_logs(&ct_kv).await {
+        log::warn!("Failed to update CT logs: {e}");
+    } else {
+        log::info!("Successfully updated CT logs");
+    }
```
mendess: nit: match would be shorter

crates/sct_validator/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+hashbrown = "0.15"
+
+# Logging
+log.workspace = true
+
+# DateTime parsing
+chrono.workspace = true
+
+# Hex encoding for debug output
+hex = "0.4"
```
mendess: can you remove all of these comments and just join and sort the dependencies into one contiguous block?

mendess (changes requested):

crates/mtc_worker/src/ccadb_roots_cron.rs
```diff
@@ -37,6 +37,23 @@ async fn main(_event: ScheduledEvent, env: Env, _ctx: ScheduleContext) {
     } else {
         log::info!("Successfully updated CCADB roots");
     }
+
+    // Update CT logs for SCT validation
```
bwesterb: If the CCADB_ROOTS_NAMESPACE is misconfigured, then we don't update CT logs for SCT validation. Is that reasonable?

crates/sct_validator/src/error.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    #[error("DER error: {0}")]
+    Der(#[from] der::Error),
+
+    #[error("SPKI error: {0}")]
+    Spki(#[from] spki::Error),
+
+    #[error("base64 decode error: {0}")]
+    Base64(#[from] base64::DecodeError),
+}
```
bwesterb: Nit: Having an error enum invites handling errors differently. Do we need the user of the crate to handle errors differently? If we're just going to log the error in any case, we don't need to expose the differences.

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+/// States that a CT log can be in.
+/// See <https://googlechrome.github.io/CertificateTransparency/ct_policy.html>
+#[derive(Clone, Debug, PartialEq, Copy, Eq)]
+pub enum LogState {
+    /// Log has been submitted but is still being evaluated.
+    Pending,
+    /// Log has passed initial evaluation period.
+    Qualified,
+    /// Log is trusted and actively accepting submissions.
+    Usable,
```
bwesterb: Qualified is also accepting submissions. The difference between Usable and Qualified, is that Qualified logs are only trusted by the newest versions of Chrome. Usable indicates that enough time has passed for a good fraction of all Chrome clients to trust the log.

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+}
+
+/// A public key used for verifying SCT signatures.
+/// CT logs use either ECDSA P-256 or RSA keys.
+#[derive(Clone, Debug)]
+pub enum VerifyingKey {
+    /// ECDSA P-256 key (most common for CT logs).
+    P256(P256VerifyingKey),
+    /// RSA key (minimum 2048 bits).
+    Rsa(RsaPublicKey),
```
bwesterb: I don't think we should bother supporting RSA. At the moment no log uses it, and I don't think any will anytime soon.

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+            current_operator,
+            previous_operators,
+        })
+    }
+
+    /// Returns the operator of this log at the given timestamp.
+    #[must_use]
+    pub fn operator_at(&self, timestamp_secs: u64) -> &str {
+        for (end_timestamp, operator) in &self.previous_operators {
+            if timestamp_secs <= *end_timestamp {
```
bwesterb: So the timestamp is inclusive: operation ends on the instant after the timestamp?

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+        }
+
+        let mut validated_scts = Vec::new();
+        let mut warnings = Vec::new();
+
+        for (index, sct) in scts.iter().enumerate() {
+            let Some(log) = self.log_list.get(&sct.log_id) else {
+                warnings.push(SctWarning::UnknownLog {
+                    index,
+                    log_id: hex::encode(sct.log_id),
```
bwesterb: Nit: probably base64 is more helpful.

crates/sct_validator/src/sct.rs
```diff
@@ -0,0 +1,150 @@
...
+    let not_before_secs = validity.not_before.to_unix_duration().as_secs();
+    let not_after_secs = validity.not_after.to_unix_duration().as_secs();
+
+    if not_after_secs < not_before_secs {
+        return Err(SctError::CertificateParse(
+            "certificate notAfter is before notBefore".to_string(),
+        ));
+    }
+
+    let lifetime_secs = not_after_secs - not_before_secs;
```
bwesterb: If notBefore = notAfter, then the certificate is valid at that instant, so one second? So should we add a `+1`? What does Chrome do?

crates/sct_validator/src/verify.rs
```diff
@@ -0,0 +1,213 @@
...
+    // Certificate length must fit in 24 bits (3 bytes)
+    let cert_len = ct_cert_der.len();
+    if cert_len > 0xFF_FFFF {
+        return Err(SctError::SignatureVerification(format!(
+            "certificate too large: {} bytes (max 16MB)",
+            cert_len
+        )));
+    }
+    let cert_len_bytes = [
+        ((cert_len >> 16) & 0xFF) as u8,
```
bwesterb: Isn't `& 0xff` superfluous here?

crates/sct_validator/src/verify.rs
```diff
@@ -0,0 +1,213 @@
...
+    data.push(SCT_VERSION_V1);
+
+    // Signature type (1 byte)
+    data.push(SIGNATURE_TYPE_CERTIFICATE_TIMESTAMP);
+
+    // Timestamp (8 bytes, big-endian)
+    data.extend_from_slice(&sct.timestamp.to_be_bytes());
+
+    // Entry type (2 bytes)
+    data.extend_from_slice(&LOG_ENTRY_TYPE_X509_ENTRY);
```
bwesterb: We'll also need support for precertificates, don't we?

bwesterb:
We should add tests with existing certificates.

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,502 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the BSD-3-Clause license found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+//! WASM-compatible SCT validation, implementing Chrome's CT policy.
+//!
+//! Chrome's policy:
```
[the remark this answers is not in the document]
lbaquerofierro: Added a note clarifying we don't enforce the legacy "one RFC 6962 log" requirement

crates/sct_validator/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+//! - Certs ≤ 180 days: 2 unique logs
+//! - Certs > 180 days: 3 unique logs
+//! - Always 2 unique operators
+//! - Logs must be Qualified/Usable/ReadOnly
+//! - Stale log list (> 70 days): auto-succeed
+//!
+//! Note: Does not enforce the legacy "one RFC 6962 log" requirement
+//! (dropped April 15th 2025 per Chrome policy update).
```
lukevalenta:
```suggestion
//! (dropped April 15th 2026 per Chrome policy update).
```

PR #173, opened 2026-01-08
mtc_worker, ct_worker: Retry requests to R2

lbaquerofierro:


crates/ct_worker/src/frontend_worker.rs
```diff
@@ -135,7 +135,11 @@ async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
                     // monitoring_url is unspecified.
                     if CONFIG.logs[name].monitoring_url.is_empty() {
                         let bucket = load_public_bucket(&ctx.env, name)?;
-                        if let Some(obj) = bucket.get(key).execute().await? {
+                        let obj_opt = with_retry(R2_MAX_RETRIES, R2_BASE_DELAY_MS, || async {
+                            bucket.get(key).execute().await
+                        })
+                        .await?;
```
bwesterb: Think here it's better to leave the retry to the client.

PR #181, opened 2026-02-11
mtc_worker: Enable logs in dashboard

cjpatton:


lukevalenta:
Could you add this change for ct_worker too?

fyi @kcxlee

PR #203, opened 2026-04-10
Add golden-file regression tests for TBS reconstruction

lukevalenta:
Add golden-file tests for the two functions that reconstruct TBSCertificate DER with an extension removed:

- extract_scts_from_cert (sct_validator): golden output is the TBS of cloudflare.pem with its SCT extension stripped
- build_precert_tbs (static_ct_api): golden output is the TBS of the precertificate in preissuer-chain.pem with CT poison stripped

The golden files capture output from the current implementation so that any future refactor of the reconstruction logic (e.g. the upcoming RustCrypto ecosystem upgrade, which rewrites these functions to work around x509-cert 0.3's private fields) is forced to produce byte-for-byte identical output.

Regenerate with: UPDATE_GOLDEN=1 cargo test -p <crate> <test_name>


lukevalenta:
Merging as this is a test-only change.

PR #205, opened 2026-04-11
Replace OnceCell::get_or_try_init with deadlock-safe non-blocking load_roots

lukevalenta:
The Workers runtime cancels any request that awaits a promise (OnceCell future) created by a different request context. Previously, a concurrent add-chain request arriving while another request was initializing ROOTS via get_or_try_init would be canceled with a 500.

Fix: load_roots now checks ROOTS.get() first (fast path), then builds the pool itself if not yet initialized, then calls ROOTS.set(). If another concurrent request races and sets first, the losing request discards its result and returns the value already in the cell. All concurrent requests do the work independently rather than waiting on each other.

Applied to both ct_worker and bootstrap_mtc_worker.


rozbb (approved):
One question: is that cancelling behavior documented anywhere? I didn’t know this was a thing to look out for

lukevalenta:
> One question: is that cancelling behavior documented anywhere? I didn’t know this was a thing to look out for

It _should_ probably be documented here: https://developers.cloudflare.com/workers/observability/errors/#the-script-will-never-generate-a-response-errors. I'll follow up with the Workers team to confirm the expected behavior and update the docs.

lukevalenta:
> > One question: is that cancelling behavior documented anywhere? I didn’t know this was a thing to look out for
> 
> It _should_ probably be documented here: https://developers.cloudflare.com/workers/observability/errors/#the-script-will-never-generate-a-response-errors. I'll follow up with the Workers team to confirm the expected behavior and update the docs.

@rozbb I submitted this PR to document the issue (pending review from the Workers team): https://github.com/cloudflare/cloudflare-docs/pull/29797

lukevalenta:
Aside from some minor fixes/rebasing, nothing substantial has changed since @rozbb 's review, so I'll go ahead and merge.

PR #206, opened 2026-04-11
Make LogEntry::Metadata a generic associated type

lukevalenta:
Adds a Metadata associated type to the LogEntry trait in tlog_tiles, allowing each tlog application to define its own sequence metadata type rather than sharing the fixed (LeafIndex, UnixTimestamp) tuple. This prepares for IETF MTC where we'll pass the old and new tree sizes from Sequencer to Frontend worker to allow for subtree calculation.

tlog_tiles:
- Adds LogEntry::Metadata: Serialize + DeserializeOwned + Copy + Default
- Adds LogEntry::make_metadata(leaf_index, timestamp, old_tree_size, new_tree_size) to construct the metadata from raw sequencer values
- TlogTilesLogEntry::Metadata = SequenceMetadata (unchanged behavior)
- SequenceMetadata type alias is preserved for backward compatibility

static_ct_api, bootstrap_mtc_api:
- Metadata = SequenceMetadata; make_metadata ignores tree sizes

generic_log_worker:
- All cache infrastructure (DedupCache, MemoryCache, CacheRead, CacheWrite) is now generic over M: the metadata type
- serialize_entries/deserialize_entries switched from fixed 32-byte binary format to serde_json (the binary format only worked for the fixed-size (u64, u64) tuple)
- PoolState<P, M>, AddLeafResult<M>, add_leaf_to_pool, sequence_entries, GenericSequencer, GenericBatcher all generified over M
- sequence_entries calls L::make_metadata(n, timestamp, old_size, new_size)
- Assert n == new_size after the sequencing loop


lukevalenta:
Close in favor of https://github.com/cloudflare/azul/pull/210

PR #204, opened 2026-04-10
Upgrade RustCrypto ecosystem to enable ml-dsa 0.1.0-rc.8

lukevalenta:
Merge https://github.com/cloudflare/azul/pull/203 first

Upgrades all RustCrypto crates to their latest RC versions to unblock ml-dsa 0.1.0-rc.8, which fixes a WASM stack overflow with ML-DSA signatures in Cloudflare Workers.

Workspace dependency changes:
- der 0.7.10 (patched fork) → 0.8.0 (upstream; Tag::RelativeOid is now native, removing the need for the fork)
- const-oid 0.9.6 → 0.10
- spki 0.7 → 0.8
- pkcs8 (new) 0.11.0-rc.11
- signature 2.2.0 → 3.0.0-rc.10
- sha2 0.10 → 0.11
- rand 0.8.5 → 0.10.0 (uses rand_core 0.10, unifying with ed25519-dalek)
- rand_core 0.6.4 → 0.10.0
- getrandom 0.2 → 0.4 (unifies with RustCrypto RC crates; 0.3 remains only in build-script deps via ahash/jsonschema, never compiled for WASM)
- ed25519-dalek 2.1.1 → 3.0.0-pre.6 (requires rand_core 0.10)
- p256 0.13 → 0.14.0-rc.8
- p384 (new) 0.14.0-rc.4
- p521 (new) 0.14.0-rc.8
- rsa 0.9 → 0.10.0-rc.17
- x509-cert 0.2.5 → 0.3.0-rc.4

Remove x509-verify and inline signature verification into x509_util, supporting P-256, P-384, P-521, and RSA PKCS#1 v1.5 (SHA-256/384/512) for TLS certificate chain validation. Ed25519 is intentionally excluded: is_link_valid validates TLS/PKI chains only; Ed25519 is not permitted by the CA/Browser Forum Baseline Requirements, does not appear in any root pool, and was never reachable. DSA, SHA-1, k256, md2, and md5 are also intentionally not supported.

Adapt to x509-cert 0.3 API changes:
- All TbsCertificate and Certificate fields are now private; replaced direct field access with getter methods (.tbs_certificate(), .subject(), .validity(), .extensions(), etc.)
- .get::<T>() renamed to .get_extension::<T>()
- Validity, Name, RelativeDistinguishedName no longer constructible via struct literal syntax; use Validity::new(), TryFrom, etc.
[...]

PR #209, opened 2026-04-13
Remove x509-verify; inline signature verification in is_link_valid

lukevalenta:
## Summary

Removes the `x509-verify` dependency and replaces it with a focused inline implementation in `is_link_valid` (`x509_util`) that supports exactly the algorithms present in real-world TLS/PKI certificate chains. This is a subset of the changes from https://github.com/cloudflare/azul/pull/204 for updating the RustCrypto ecosystem (x509-verify uses but is not part of RustCrypto, and is not kept up-to-date with release candidate versions).

**Supported algorithms:**
- ECDSA P-256 (`ecdsa-with-SHA256`)
- ECDSA P-384 (`ecdsa-with-SHA384`)
- ECDSA P-521 (`ecdsa-with-SHA512`) (notably, was not enabled in the existing `x509-verify` feature set)
- RSA PKCS#1 v1.5 (SHA-1, SHA-256, SHA-384, SHA-512)

**Intentionally excluded** (none appear in any configured root pool, and none were enabled in our existing `x509-verify` feature set):
- Ed25519 (not permitted in TLS by CA/Browser Forum Baseline Requirements)
- DSA, MD2, MD5, k256, P-192, P-224

**Other changes:**
- Add `p384` and `p521` as an explicit workspace dependencies (previously only transitive deps via `x509-verify`)
- Replace `x509_verify::spki::Error` with `spki::Error` directly in `static_ct_api::StaticCTError` — `x509-verify` was only re-exporting `spki`
- Remove `x509-verify` from `integration_tests` and from the `cargo-machete` ignored list; also remove `signature` from that list (was never directly used)


crates/x509_util/src/lib.rs
```diff
@@ -431,26 +430,88 @@ fn check_well_formedness(cert: &Certificate) -> Result<(), ValidationError> {
...
+        return false;
+    };
+    let Ok(spki_der) = issuer.tbs_certificate.subject_public_key_info.to_der() else {
+        return false;
+    };
+
+    match child.signature_algorithm.oid {
+        // ecdsa-with-SHA256 → P-256; ecdsa-with-SHA384/SHA512 → P-384.
+        ECDSA_WITH_SHA_256 => verify_p256(&spki_der, &tbs_der, sig_bytes),
+        ECDSA_WITH_SHA_384 | ECDSA_WITH_SHA_512 => verify_p384(&spki_der, &tbs_der, sig_bytes),
```
Lekensteyn: Shouldn't ECDSA_WITH_SHA_512 be paired with P521?

Analysis based on ca-certificates 2026-03-19 covering Mozilla's trust store:
```
$ openssl x509 -multi -in /opt/homebrew/share/ca-certificates/cacert.pem -text | awk '/^    Signature Algorithm:/{print $3}'|sort|uniq -c|sort -n

   1 ecdsa-with-SHA512 <-- improperly supported?
   4 sha512WithRSAEncryption
   7 ecdsa-with-SHA256
  18 sha1WithRSAEncryption <-- purposely not supported because it is legacy?
  20 sha384WithRSAEncryption
  36 ecdsa-with-SHA384
  59 sha256WithRSAEncryption
```

The above ecdsa-with-SHA512 signature comes from this certificate with a P-521 public key:
* Serial Number: e8:6f:18:7b:d6:39:6b:98:4a:49:98:0a
* Subject: C=HU, L=Budapest, O=Microsec Ltd., organizationIdentifier=VATHU-23584497, CN=e-Szigno TLS Root CA 2023
* Not Before: Jul 17 14:00:00 2023 GMT
* Not After : Jul 17 14:00:00 2038 GMT

Can we have tests with these real-world examples?

lukevalenta: > Shouldn't ECDSA_WITH_SHA_512 be paired with P521?
Err, yes good catch. We should either add p521 support or drop the SHA512 from the match.

> 18 sha1WithRSAEncryption
The CA/B forum requirements say SHA1 is not recommended but support is grandfathered in, but it looks like we should support it!

Thanks for the analysis, Peter. I'll update the support and add test with these certs.

lukevalenta: Done. The new tests currently use synthetic example chains, but happy to try to track down some real-world chains if you think that would be better!

crates/x509_util/src/lib.rs
```diff
@@ -431,26 +430,112 @@ fn check_well_formedness(cert: &Certificate) -> Result<(), ValidationError> {
...
+        SHA_256_WITH_RSA_ENCRYPTION => verify_rsa::<sha2::Sha256>(&spki_der, &tbs_der, sig_bytes),
+        SHA_384_WITH_RSA_ENCRYPTION => verify_rsa::<sha2::Sha384>(&spki_der, &tbs_der, sig_bytes),
+        SHA_512_WITH_RSA_ENCRYPTION => verify_rsa::<sha2::Sha512>(&spki_der, &tbs_der, sig_bytes),
+        _ => false,
     }
 }
 
+fn verify_p256(spki_der: &[u8], tbs_der: &[u8], sig_bytes: &[u8]) -> bool {
+    use signature::Verifier as _;
+    let Ok(spki) = spki::SubjectPublicKeyInfoRef::try_from(spki_der) else {
```
rozbb: Nit: I think you can refactor this line into the caller. Saves 5x repetition

rozbb (approved):
LGTM! I got a little confused bc your first comment said that P-521 and SHA1 are excluded, but they are in fact supported

lukevalenta:
> LGTM! I got a little confused bc your first comment said that P-521 and SHA1 are excluded, but they are in fact supported

Oops, I've fixed the description!

crates/x509_util/src/lib.rs
lukevalenta: Good call -- done! I also moved the repeated `use signature::Verifier as _;` imports to the top of the file.

lukevalenta:
I'm going to go ahead and merge to unblock other changes, but happy to address any other feedback in followup PRs!

PR #210, opened 2026-04-13
Make LogEntry::Metadata a generic associated type

lukevalenta:
Adds a Metadata associated type to the LogEntry trait in tlog_tiles,
allowing each tlog application to define its own sequence metadata type
rather than sharing the fixed (LeafIndex, UnixTimestamp) tuple.

tlog_tiles:
- Adds LogEntry::Metadata: Serialize + DeserializeOwned + Copy + Default
- Adds LogEntry::make_metadata(leaf_index, timestamp, old_tree_size,
  new_tree_size) to construct the metadata from raw sequencer values
- TlogTilesLogEntry::Metadata = SequenceMetadata (unchanged behavior)
- SequenceMetadata type alias is preserved for backward compatibility

static_ct_api, bootstrap_mtc_api:
- Metadata = SequenceMetadata; make_metadata ignores tree sizes

generic_log_worker:
- All cache infrastructure (DedupCache, MemoryCache, CacheRead,
  CacheWrite) is now generic over M: the metadata type
- serialize_entries/deserialize_entries switched from fixed 32-byte
  binary format to serde_json (the binary format only worked for the
  fixed-size (u64, u64) tuple)
- PoolState<P, M>, AddLeafResult<M>, add_leaf_to_pool, sequence_entries,
  GenericSequencer, GenericBatcher all generified over M
- sequence_entries calls L::make_metadata(n, timestamp, old_size, new_size)
- Assert n == new_size after the sequencing loop
- CheckpointCallbacker gains old_tree_size and new_tree_size parameters
  so callbacks can identify which subtrees were covered by each batch

### DIFF
--- a/crates/generic_log_worker/src/batcher_do.rs
+++ b/crates/generic_log_worker/src/batcher_do.rs
@@ -7,9 +7,10 @@
 //! Entries are assigned to Batcher shards with consistent hashing on the cache key.
 
 use crate::{
-    deserialize, get_durable_object_stub, load_cache_kv, obs, serialize, LookupKey,
-    SequenceMetadata, BATCH_ENDPOINT, ENTRY_ENDPOINT, SEQUENCER_BINDING,
+    deserialize, get_durable_object_stub, load_cache_kv, obs, serialize, CacheSerialize,
+    LookupKey, BATCH_ENDPOINT, ENTRY_ENDPOINT, SEQUENCER_BINDING,
 };
+use serde::{de::DeserializeOwned, Serialize};
 use base64::prelude::*;
 use futures_util::future::{join_all, select, Either};
 use std::{
@@ -23,12 +24,17 @@ use worker::kv::KvStore;
 #[allow(clippy::wildcard_imports)]
 use worker::*;
 
-pub struct GenericBatcher {
+/// A Durable Object that buffers incoming log entries and submits them to the
+/// Sequencer in batches.
+///
+/// `M` is the sequence metadata type produced for each entry after sequencing
+/// (i.e., [`LogEntry::Metadata`](tlog_tiles::LogEntry::Metadata)).
+pub struct GenericBatcher<M> {
     env: Env,
     config: BatcherConfig,
     state: State,
     kv: Option<KvStore>,
-    batch: RefCell<Batch>,
+    batch: RefCell<Batch<M>>,
     in_flight: RefCell<usize>,
     processed: RefCell<usize>,
     wshim: Option<obs::Wshim>,
@@ -42,14 +48,15 @@ pub struct BatcherConfig {
     pub location_hint: Option<String>,
 }
 
-// A batch of entries to be submitted to the Sequencer together.
-struct Batch {
+// A batch of entries to be submitted to the Sequencer together. `M` is the
+// sequence metadata type returned to waiters once the batch is sequenced.
+struct Batch<M> {
     entries: Vec<PendingLogEntryBlob>,
     by_hash: HashSet<LookupKey>,
-    done: Sender<HashMap<LookupKey, SequenceMetadata>>,
+    done: Sender<HashMap<LookupKey, M>>,
 }
 
-impl Default for Batch {
+impl<M: Clone + Default> Default for Batch<M> {
     /// Returns a batch initialized with a watch channel.
     fn default() -> Self {
         Self {
@@ -60,7 +67,7 @@ impl Default for Batch {
     }
 }
 
-impl GenericBatcher {
+impl<M: CacheSerialize + Serialize + DeserializeOwned + Clone + Default + Copy + 'static> GenericBatcher<M> {
     /// Returns a new batcher with the given config.
     ///
     /// # Panics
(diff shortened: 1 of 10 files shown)


lbaquerofierro (commented):
Nice change overall. The backward compatibility is really well done

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -863,8 +869,10 @@ async fn sequence_entries<L: LogEntry>(
             }
         }
 
-        // Add the entry and metadata to our lists of things sequenced
-        let metadata = (n, timestamp);
+        // Add the entry and metadata to our lists of things sequenced.
+        // L::make_metadata provides the application-specific metadata type,
```
lbaquerofierro, line 873: I think there might be an issue with the tree size semantics though. Looking at log_ops.rs around line 873, when you call make_metadata(n, timestamp, old_size, new_size), wouldn't each entry think the tree jumps from old_size directly to the final size?
Like entry 0 sees the tree go from old_size to (old_size + N), but entry 1 also sees the same jump, even though it's actually being added when the tree is at old_size + 1.
Maybe the new_tree_size param should be n instead? Or if it's intentionally the batch end size, maybe add a comment clarifying that?

crates/generic_log_worker/src/lib.rs
```diff
@@ -401,40 +416,93 @@ impl DedupCache {
...
+    /// The `String` can be converted to a `worker::Error` via `From<String>`.
+    fn deserialize_entries(buf: &[u8]) -> std::result::Result<Vec<(LookupKey, Self)>, String>;
 }
 
-fn deserialize_entries(buf: &[u8]) -> Result<Vec<(LookupKey, SequenceMetadata)>> {
-    if !buf.len().is_multiple_of(32) {
-        return Err("invalid buffer length".into());
+/// `SequenceMetadata` uses the original 32-byte binary format for backward
+/// compatibility with existing Durable Object storage.
+impl CacheSerialize for SequenceMetadata {
```
lbaquerofierro, line 443: nit: since the SequenceMetadata binary format is so critical for backward compat (nice regression test btw), maybe a warning comment on the CacheSerialize impl could be useful (Something about how changing the format would corrupt existing DO storage)

crates/generic_log_worker/src/log_ops.rs
lukevalenta: Ah yes, we want the `new_size` in the metadata to be the batch end size, since the `[old_tree_size, new_tree_size)` range is what is used to construct the subtrees. I'll add a comment to clarify!

lukevalenta: So just to clarify further, all of the entries in the batch get sequenced together, and their leaf indexes will be in the range `[old_size, new_size)`. For standalone MTCs, the inclusion proof for the individual entries goes to one of the two subtrees of the `[old_size, new_size)` range.

crates/generic_log_worker/src/lib.rs
lukevalenta: Done!

crates/generic_log_worker/src/batcher_do.rs
```diff
@@ -23,12 +24,12 @@ use worker::kv::KvStore;
 #[allow(clippy::wildcard_imports)]
 use worker::*;
 
-pub struct GenericBatcher {
+pub struct GenericBatcher<M> {
```
rozbb, line 32: Would be good to comment what M is. Or even better the whole struct if you feel like it

crates/generic_log_worker/src/batcher_do.rs
```diff
@@ -43,13 +44,13 @@ pub struct BatcherConfig {
 }
 
 // A batch of entries to be submitted to the Sequencer together.
-struct Batch {
+struct Batch<M> {
```
rozbb, line 53: Ditto

crates/generic_log_worker/src/lib.rs
```diff
@@ -401,40 +415,98 @@ impl DedupCache {
...
+    M::serialize_entries(entries)
+}
+
+fn deserialize_entries<M: CacheSerialize>(buf: &[u8]) -> Result<Vec<(LookupKey, M)>> {
+    M::deserialize_entries(buf).map_err(Into::into)
 }
 
 // A fixed-size in-memory FIFO cache.
-struct MemoryCache {
+struct MemoryCache<M> {
```
rozbb, line 504: Ditto

crates/generic_log_worker/src/log_ops.rs
```diff
@@ -918,14 +927,16 @@ async fn sequence_entries<L: LogEntry>(
         }
     }
 
+    assert_eq!(n, new_size, "loop must have processed exactly entries.len() entries");
+
     // Stage leftover partial data tile, if any.
-    if n != old_size && n % u64::from(TlogTile::FULL_WIDTH) != 0 {
+    if new_size != old_size && !new_size.is_multiple_of(u64::from(TlogTile::FULL_WIDTH)) {
```
rozbb, line 933: TIL

rozbb (approved):

PR #216, opened 2026-04-17
Decouple sequencer metadata from LogEntry via per-worker SequencerMetadata types

lukevalenta:
This PR refactors some of the changes introduced in https://github.com/cloudflare/azul/pull/210 for cleaner API boundaries, and setting up for the IETF MTC worker which will need to pass the old and new tree sizes from Sequencer to Frontend Worker in addition to the existing leaf index and timestamp fields.

Motivation: cleaner API boundaries between spec-level and Workers-level
crates. Previously tlog_tiles (a published crate implementing the public
c2sp.org/tlog-tiles spec) carried a Metadata associated type on LogEntry
that existed only to describe Cloudflare Durable Object dedup cache wire
formats -- a concern of the sequencer/batcher/frontend runtime, not of the
tlog spec. Moving this concept into generic_log_worker keeps tlog_tiles
focused on the spec and lets each worker own the metadata shape (and wire
format) that matches its own deployment constraints. AGENTS.md is updated
to codify this boundary.

* Decouple sequencer metadata from LogEntry via per-worker SequencerMetadata types

  Introduces a SequencerMetadata trait in generic_log_worker that captures the
  sequencer-level responsibilities previously attached to LogEntry::Metadata:

    - constructing metadata for a sequenced entry (make)
    - serializing/deserializing it in the short-term dedup cache ring buffer
      stored in Durable Object storage (serialize_cache_entries /
      deserialize_cache_entries; defaults use serde_json)

  Values of a SequencerMetadata type also flow over the wire as the per-entry
  response from the Sequencer DO to the Batcher DO and on to the Worker
  frontend (via bitcode), and -- for ct_worker -- back out as JSON metadata in
  the long-term KV dedup cache. Each worker's SequencerMetadata type must
  preserve these wire formats across upgrades.

  LogEntry no longer has a Metadata associated type; GenericSequencer is now
  generic over <L: LogEntry, M: SequencerMetadata>, and GenericBatcher is
  generic over <M: SequencerMetadata> alone (it never inspects entry contents).
  The CacheSerialize trait and impl_json_cache_serialize! macro are removed.

  Each worker now owns its metadata type as a local tuple struct newtype over
  (LeafIndex, UnixTimestamp):

    - ct_worker::StaticCTSequenceMetadata preserves both the 32-byte binary DO
      cache format and the `[leaf_index, timestamp]` JSON format used by the
      long-term KV dedup cache (automatic for tuple structs under serde).

    - bootstrap_mtc_worker::BootstrapMtcSequenceMetadata preserves the same
      32-byte binary DO cache format so existing ring-buffer entries continue
      to deserialize after deploy. (Bootstrap does not use the long-term KV
      cache, so no JSON wire format constraint applies.)

  The shared binary format lives in generic_log_worker as the free functions
  serialize_sequence_metadata_entries / deserialize_sequence_metadata_entries,
  with a regression test pinning the 32-byte layout.

* Simplify LogEntry::new to take (leaf_index, timestamp) instead of Metadata

  The metadata argument was only used to pass leaf_index and timestamp into the
  sequenced entry struct; only StaticCTLogEntry actually read the fields. Taking
  these two values directly removes a layer of indirection and decouples
  LogEntry::new from the Metadata associated type, paving the way for separating
  sequencer metadata from entry-construction metadata.

### DIFF
--- a/crates/generic_log_worker/src/lib.rs
+++ b/crates/generic_log_worker/src/lib.rs
@@ -26,11 +26,9 @@ use std::cell::RefCell;
 use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, HashMap, VecDeque};
 
-use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use serde::de::DeserializeOwned;
-use std::io::Write;
 pub use tlog_tiles::LookupKey;
-use tlog_tiles::{PendingLogEntry, SequenceMetadata};
+use tlog_tiles::{LeafIndex, PendingLogEntry, UnixTimestamp};
 use tokio::sync::Mutex;
 use util::now_millis;
 use worker::{
@@ -227,7 +225,7 @@ struct DedupCache<M> {
     storage: Storage,
 }
 
-impl<M: CacheSerialize + Serialize + DeserializeOwned + Clone + Copy + Default + 'static> CacheWrite<M> for DedupCache<M> {
+impl<M: SequencerMetadata> CacheWrite<M> for DedupCache<M> {
     /// Write entries to both the short-term deduplication cache and its backup in DO Storage.
     async fn put_entries(&self, entries: &[(LookupKey, M)]) -> Result<()> {
         if entries.is_empty() {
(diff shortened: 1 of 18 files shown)


crates/generic_log_worker/src/lib.rs
```diff
@@ -415,88 +413,129 @@ impl<M: CacheSerialize + Serialize + DeserializeOwned + Clone + Copy + Default +
...
     ///
-    /// # Errors
+    /// Every log entry has a `leaf_index` and sequencing `timestamp`; these two
+    /// are always populated. `old_tree_size` and `new_tree_size` are the tree
+    /// sizes immediately before and after the sequencing batch that included
+    /// this entry. They are unused by the default `SequenceMetadata` shape but
+    /// are exposed here so log-specific metadata types can carry them (for
+    /// example, an IETF MTC worker uses them to identify the covering subtree
+    /// for an inclusion proof without enumerating candidates).
+    fn make(
```
rozbb: Nit: convention is to call this `new`

rozbb (approved):

PR #217, opened 2026-04-17
AGENTS.md: document pre-push checks; add advisory rustfmt and machete CI steps

lukevalenta:
## Summary

- Expand the \"Always run before pushing\" boundary in AGENTS.md into a concrete pre-push checklist (clippy, tests, fmt, machete), so contributors (and agents) can verify their work against the same checks CI runs
- Document the preference for fixup commits (`git commit --fixup=<sha>` + `git rebase --autosquash`) over layered \"fix lint\" commits when a pre-push check surfaces an issue
- Add `cargo fmt --all --check` and `cargo machete` as **advisory** CI steps (`continue-on-error: true`, matching the existing pedantic-clippy step). Formatting and unused-dependency issues should be fixed eagerly, but they don't warrant blocking a merge

## Motivation

Recent PR work surfaced a `cargo machete` violation (a speculative `signature` dep in `integration_tests/Cargo.toml` that was never used) and noisy rustfmt drift in some files. Making both advisory in CI raises visibility without turning routine housekeeping into merge blockers, while the AGENTS.md checklist makes it easy for an agent reviewing its own work to catch these before pushing.

## Test plan

- \`cargo fmt --all --check\` and \`cargo machete\` both run locally against the workspace
- The CI steps use \`continue-on-error: true\` so they surface findings without failing the required \"Lint\" job (same pattern as the existing \"Clippy (pedantic, advisory)\" step)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,8 +20,12 @@ jobs:
       - name: Clippy (pedantic, advisory)
         run: cargo clippy --tests -- -W clippy::pedantic
         continue-on-error: true
-      - name: Unused dependencies
+      - name: Unused dependencies (advisory)
         run: cargo install cargo-machete --quiet && cargo machete
+        continue-on-error: true
+      - name: Formatting (advisory)
+        run: cargo fmt --all --check
+        continue-on-error: true
 
   build:
     name: Build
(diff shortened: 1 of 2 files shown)


lukevalenta:
This is a simple change to add extra advisory lints and update AGENTS.md. Merging.

PR #218, opened 2026-04-18
Add tlog_witness crate and witness_worker implementing c2sp.org/tlog-witness

lukevalenta:
Introduces two new crates and an integration-test suite.

tlog_witness (spec-level, publishable to crates.io)

- `AddCheckpointRequest` + parsers/serializers for the `POST /add-checkpoint`
  body format (`old <N>\n<proof-lines>\n\n<checkpoint-note>`) in both
  directions (log → witness requests, witness → log responses).
- `MAX_CONSISTENCY_PROOF_LINES = 63`, matching the spec.
- `CONTENT_TYPE_TLOG_SIZE = "text/x.tlog.size"` for 409 response bodies.
- `TlogWitnessError` with variants for malformed request / response / note.
- 12 unit tests covering happy path, round-trips, spec-example shape, and
  malformed-input rejection.

witness_worker (Cloudflare Worker deployment)

- `POST /add-checkpoint` implementing all the MUSTs from the spec:
  - Look up log by origin → 404 on unknown.
  - Verify the checkpoint note via `signed_note::Note::verify`. Both
    `UnverifiedNote` (no trusted-key signature at all) and
    `InvalidSignature` (a trusted-key signature that failed to verify,
    making the note malformed per c2sp.org/signed-note) are surfaced as
    403 Forbidden, matching sunlight and sigsum-go. The specific
    `NoteError` variant is logged for debugging; other variants are
    surfaced as 400 Bad Request.
  - Range-check `old_size <= checkpoint.size` → 400 on violation.
  - Single-RPC atomic check-proof-and-update against the per-origin
    `WitnessState` Durable Object. The DO compares `old_size` against
    stored state (409 with `text/x.tlog.size` body on mismatch or
    same-size-different-hash), verifies the Merkle consistency proof
    against its stored `latest_hash` (422 on failure), and writes the
    new `(size, hash)`. Atomicity comes from Durable Objects' input and
    output gates, documented inline at the handler; verifying the proof
    inside the same handler that reads and writes state keeps the whole
    sequence inside one atomic transaction and one DO round trip.
  - On 200 from the DO, produce and return a `cosignature/v1` signature.
- `GET /metadata` returns the witness's identity (name + DER SPKI),
  description, submission/monitoring URL prefixes, and the full list of
  configured logs (origin + trusted public keys). `description` fields
  are `Option<&str>` with `#[serde(skip_serializing_if = "Option::is_none")]`
  so they are omitted (not `null`) when unset; regression tests pin this.
- `WitnessState` Durable Object (one per log origin, keyed by origin) holds
  per-origin `(latest_size, latest_hash)` state. The DO's on-disk JSON
  layout and the internal RPC wire shape (including the embedded
  consistency-proof hashes) are pinned by regression tests so format drift
  is caught before it can orphan deployed state.
- Config model (per-log `origin` + `log_public_keys` SPKI list +
  top-level `witness_name`, `submission_prefix`, `monitoring_prefix`;
  `WITNESS_SIGNING_KEY` secret) validated at build time against a JSON
  Schema.
- Produces `cosignature/v1` signatures using the existing
  `tlog_tiles::CosignatureV1CheckpointSigner` (Ed25519-only, matching the
  witness spec's algorithm requirement).

Integration tests

- New `crates/integration_tests/tests/tlog_witness.rs` drives the witness
  end-to-end against a live `wrangler dev` instance: `/metadata`, two
  successful `add-checkpoint` submissions (with cosignature verification
  against the metadata-advertised public key), then 409 / 404 / 403 / 422 /
  400 error cases — including the trusted-(name,id)-with-bad-signature
  case that verifies the 403-on-InvalidSignature behavior. The cases are
  sequenced inside a single `#[tokio::test]` threaded through one
  in-memory `ToyLog` so the witness's persisted state evolves
  predictably and consistency proofs produced by the test always match
  what the witness expects.
- New `integration-tlog-witness` CI job in
  `.github/workflows/integration.yml` builds the worker, starts
  `wrangler dev`, waits for `/metadata` to respond, then runs the test
  suite.
- `crates/witness_worker/config.dev.json` now ships an Ed25519 SPKI whose
  matching seed is embedded in the test source as `LOG_SIGNING_KEY_PEM`.
  This keypair is dev-only, clearly documented as such, and never used in
  production.

This witness is deployable independently of any MTC work and can cosign
checkpoints from any checkpoint-based log (static-ct, MTC, sigsum, etc.).


crates/tlog_witness/src/add_checkpoint.rs
```diff
@@ -0,0 +1,405 @@
...
+//! <proof-line-2>\n
+//! ...
+//! \n                   ← empty line
+//! <checkpoint note>    ← a c2sp.org/signed-note checkpoint, terminated by its signature block
+//! ```
+//!
+//! The response body (on HTTP 200) is a sequence of c2sp.org/signed-note
+//! signature lines, each beginning with `—` (U+2014) and ending with `\n`.
+//!
+//! Unlike the MTC §C.2 `sign-subtree` body (which has multiple note-like
```
lukevalenta: Remove this part about sign-subtree as it's not relevant to the tlog-witness spec.

crates/witness_worker/src/frontend_worker.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/// Top-level `#[event(fetch)]` handler.
+#[event(fetch, respond_with_errors)]
+async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
+    Router::new()
+        .post_async("/add-checkpoint", |req, ctx| async move {
+            add_checkpoint(req, ctx.env).await
+        })
+        .get("/", |_req, _ctx| {
```
lukevalenta, line 58: Let's also add a /metadata endpoint (similar to the bootstrap_mtc_worker) that displays the witness name, description, public key, and two URL prefixes: the submission prefix for write APIs and the monitoring prefix for read APIs. Note that the monitoring prefix is currently unused in the spec. The metadata endpoint can also include the full list of configured logs.

crates/witness_worker/src/frontend_worker.rs
```diff
@@ -0,0 +1,328 @@
...
+            return Response::error("Bad request: checkpoint note has no origin", 400);
+        }
+    };
+    let Some(log) = CONFIG.log_by_origin(origin) else {
+        return Response::error("Unknown log origin", 404);
+    };
+
+    // (3) Verify the checkpoint signature against trusted log keys.
+    //
+    // tlog-witness semantics differ slightly from `open_checkpoint`: the
```
lukevalenta: Perhaps we update open_checkpoint to accept an enum (ValidationMode::{Any, All}) that describes how validation should be done?

crates/witness_worker/src/frontend_worker.rs
```diff
@@ -0,0 +1,328 @@
...
+    // witness only needs AT LEAST ONE valid signature from a trusted key
+    // (and MUST ignore unknown-key signatures), whereas `open_checkpoint`
+    // requires signatures from every verifier in the list. So we drive
+    // [`Note::verify`] directly.
+    let now_millis = u64::try_from(
```
lukevalenta: Use the generic_log_worker::now_millis() util function: https://github.com/cloudflare/azul/blob/3c97cfc609b287f2a088b204cd119efc12f560ff/crates/generic_log_worker/src/util.rs#L14

crates/witness_worker/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+//! client (typically the log itself) submits a new checkpoint along with a
+//! consistency proof from the witness's latest recorded state, and the
+//! witness returns a timestamped `cosignature/v1` signature.
+//!
+//! Per-log state (the latest cosigned tree size and root hash) is persisted
+//! in a [`WitnessState`] Durable Object, one per log origin. The DO's
+//! single-threaded execution model provides the atomic "check-old-size,
+//! update-latest, return-cosignature" sequence that the spec requires.
+//!
+//! A witness is not a full log mirror: it does not store tiles or serve log
```
lukevalenta: We can remove this paragraph about sign-subtrees and mirrors.

crates/witness_worker/src/witness_state_do.rs
```diff
@@ -0,0 +1,174 @@
...
+                // hashes. The frontend enforces this before reaching us, but
+                // double-check defensively.
+                if body.old_size == body.new_size && current.hash.0 != body.new_hash.0 {
+                    return Response::from_json(&current).map(|r| r.with_status(409));
+                }
+                let new_state = LatestCheckpoint {
+                    size: body.new_size,
+                    hash: body.new_hash,
+                };
+                self.state.storage().put(STATE_KEY, &new_state).await?;
```
lukevalenta, line 180: Document why input/output gates make this get-and-put operation atomic and safe.

crates/witness_worker/README.md
```diff
@@ -0,0 +1,49 @@
+# Transparency Log Witness Worker
+
+A [`c2sp.org/tlog-witness`](https://c2sp.org/tlog-witness) implementation on
+[Cloudflare Workers](https://workers.cloudflare.com/).
+
+A *witness* is a non-CA service that cosigns transparency-log checkpoints
```
lukevalenta: Remove the reference to CA, as witnesses are more general.

crates/witness_worker/README.md
```diff
@@ -0,0 +1,49 @@
...
+[`tlog_witness`](../tlog_witness/) crate (published to crates.io); this crate
+is the Cloudflare Workers-specific shell that wires them up to HTTP, Durable
+Object storage for per-origin state, and a secret-managed Ed25519 signing
+key.
+
+## Known limitations
+
+- Only Ed25519 logs are supported (the only algorithm for `cosignature/v1`
+  per c2sp.org/tlog-cosignature v1).
+- No `/sign-subtree` endpoint yet; that will be added in a follow-up commit
```
lukevalenta: We can remove this too, as save it for the mirror implementation.

crates/witness_worker/config.schema.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "description": {
+            "type": "string"
+        },
+        "logs": {
+            "type": "object",
+            "description": "Logs this witness will accept checkpoints for, keyed by a short identifier used only internally. The identifier does not appear on the wire.",
+            "additionalProperties": {
+                "type": "object",
+                "required": ["origin", "log_public_keys_spki_b64"],
```
lukevalenta: Lets call this just log_public_keys

crates/witness_worker/README.md
```diff
@@ -0,0 +1,46 @@
...
+The wire-format parsers and serializers live in the
+[`tlog_witness`](../tlog_witness/) crate (published to crates.io); this crate
+is the Cloudflare Workers-specific shell that wires them up to HTTP, Durable
+Object storage for per-origin state, and a secret-managed Ed25519 signing
+key.
+
+## Known limitations
+
+- Only Ed25519 logs are supported (the only algorithm for `cosignature/v1`
+  per c2sp.org/tlog-cosignature v1).
```
lbaquerofierro: Small wording thing: this reads as if the Ed25519-only log-key constraint follows from cosignature/v1 being Ed25519, but they're two independent things, cosignature/v1 forces the cosigning key to be Ed25519, while accepting only Ed25519 log keys is a separate verification-side choice. Maybe split into two bullets:
- Cosignatures are Ed25519 only (per c2sp.org/tlog-cosignature v1).
- Only logs that sign with Ed25519 are accepted.

lbaquerofierro (approved):

crates/witness_worker/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Values are plain `(VerifyingKey, key_id)` pairs rather than a
+/// pre-built `VerifierList`, because `Box<dyn NoteVerifier>` is not
+/// `Sync` and so cannot live inside a `LazyLock`. Building the
+/// `VerifierList` per request from these cached keys is cheap
+/// (`Ed25519NoteVerifier::new` is just field assignment) and keeps the
+/// startup-time validation entirely in this module.
+pub(crate) static LOG_KEYS: LazyLock<HashMap<String, Vec<LogKey>>> = LazyLock::new(|| {
+    let mut out = HashMap::with_capacity(CONFIG.logs.len());
+    for (id, log) in &CONFIG.logs {
+        out.insert(log.origin.clone(), parse_log_keys(id, log));
```
lbaquerofierro: Small thing on origin uniqueness: if two entries in `config.logs` end up with the same `origin`, `HashMap::insert` will quietly overwrite and only the last entry's keys stick around. Not a security issue (anything signed by the dropped keys just 403s), more of a config footgun where you'd see "why aren't my signatures verifying" without an obvious cause.

Could mirror the `(name, key_id)` collision check that's already in `parse_log_keys`, something like:
```rust
assert!(out.insert(log.origin.clone(), parse_log_keys(id, log)).is_none(),
    "duplicate origin {:?} across log entries", log.origin);
```

crates/witness_worker/config.schema.json
```diff
@@ -0,0 +1,52 @@
...
+        "logs": {
+            "type": "object",
+            "description": "Logs this witness will accept checkpoints for, keyed by a short identifier used only internally. The identifier does not appear on the wire.",
+            "additionalProperties": {
+                "type": "object",
+                "required": ["origin", "log_public_keys"],
+                "additionalProperties": false,
+                "properties": {
+                    "description": {"type": "string"},
+                    "origin": {
```
lbaquerofierro: `witness_name` has `"pattern": "^[^+\\s]+$"` to match signed-note's KeyName rules, but `origin` ends up with the same constraint at runtime and there's no pattern check on it here. A typo won't fail schema validation or a deployment, but the first `/add-checkpoint` request will panic when `LOG_KEYS` lazy-initializes. It could mirror the `witness_name` pattern on `origin` to catch this at build time instead.

lukevalenta: Added a new commit that reworks the config a bit to address your feedback!

PR #219, opened 2026-04-20
Fix cargo fmt and clippy complaints, and use latest rust 1.95 toolchain in CI

lukevalenta:
Fix cargo fmt and clippy complaints, and use latest rust 1.95 toolchain in CI

Also remove "fuzz" from workspace.default-members so that `cargo test`
runs unit tests without also building the cargo-fuzz harness (which depends
on libfuzzer-sys and a nightly sanitizer runtime).  Fuzz targets are still
linted by `cargo clippy --workspace --all-targets` and run explicitly with
`cargo fuzz run <target>`.

### DIFF
--- a/.github/workflows/bonk.yml
+++ b/.github/workflows/bonk.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 1
-
+      - run: rustup update stable && rustup default stable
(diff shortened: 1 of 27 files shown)


mendess (approved):

lukevalenta:
Made a couple more tweaks to actually address all of the advisory lints and fmts. Merging.

PR #220, opened 2026-04-20
Update bonk to opus 4.7

lukevalenta:


lukevalenta:
Simple non-production change. Merging to see if it fixes the `ProviderInitError`s.

PR #221, opened 2026-04-21
Add tlog_subtree_signature crate with sign-subtree wire-format and binary-format support

lukevalenta:
Introduces a new spec-level crate, separate from `ietf_mtc_api`, for the
`sign-subtree` cosigner endpoint. The endpoint is currently specified in
draft-ietf-plants-merkle-tree-certs-02 §C.2; the intent is for it to
migrate into a standalone C2SP specification (provisional URL
`c2sp.org/tlog-subtree-signature`), and this crate is positioned to ship
with that future spec.

Scope:

  HTTP wire format (top-level module):

  Server side (cosigner receiving the request):
  - `SignSubtreeRequest`, `SubtreeNoteBody`
  - `parse_sign_subtree_request`
  - `serialize_sign_subtree_response`

  Client side (requester):
  - `serialize_subtree_note_body`
  - `serialize_sign_subtree_request`
  - `parse_sign_subtree_response`

  Shared:
  - `MAX_CONSISTENCY_PROOF_HASHES = 63`
  - `TlogSubtreeSignatureError`

  Binary signing format (`crypto` module, re-exported at crate root):

  - `serialize_subtree_signature_input` — produce the `mtc-subtree/v1\n\0`
    binary message a subtree signer actually signs (§5.4.1). Opaque-bytes
    interface: `cosigner_id` and `log_id` are `&[u8]` so the crate doesn't
    depend on any particular identity encoding (MTC uses BER-encoded
    `TrustAnchorID`; future C2SP spec may use something else).
  - `RawSigner` / `RawVerifier` traits: algorithm-agnostic abstraction
    over "produce raw sig bytes for a message" / "check raw sig bytes for
    a message". Concrete implementations (Ed25519, ML-DSA-44, ...) live
    in downstream crates; today the only workspace ones are in
    `ietf_mtc_api` (`MtcSigningKey` / `MtcVerifyingKey`).
  - `sign_subtree(signer, cosigner_id, log_id, start, end, hash)` —
    convenience wrapper that builds the binary input and hands it to the
    signer.
  - `SubtreeNoteVerifier<V: RawVerifier>` — a `signed_note::NoteVerifier`
    that accepts a subtree signed note, reconstructs the binary signing
    input from the note text body, and delegates to the wrapped
    `RawVerifier`. Caller supplies `KeyName`, key ID, and identity bytes;
    this crate does not prescribe name-format or key-ID conventions.

The label `mtc-subtree/v1` is kept as-is for wire compatibility with
current IETF MTC deployments. When the C2SP spec settles on a new label
(e.g. via [C2SP#237]'s `subtree/v1`), `serialize_subtree_signature_input`
will gain a second encoding and existing clients can migrate in lockstep.

Out of scope (stays in `ietf_mtc_api`):

  - Multi-algorithm key enums (`MtcSigningKey` / `MtcVerifyingKey`). They
    become `RawSigner` / `RawVerifier` impls in a follow-up commit.
  - `TrustAnchorID` / `ID_RDNA_TRUSTANCHOR_ID` and the MTC `oid/…`
    key-name scheme. The new crate takes an opaque `KeyName` + key ID.
  - `MtcCosigner` and `ParsedMtcProof`. These remain MTC-CA-specific.

Tests: 21 unit tests (17 wire-format + 4 binary-format) pin the request /
response round-trip, malformed-input rejection, the `mtc-subtree/v1`
binary layout, and the `SubtreeNoteVerifier` happy-path + malformed-text
paths. The binary-format layout test mirrors the pattern used by
`SequenceMetadata::serialize_cache_entries` — any change to the format
would fail loudly and visibly.

The `parse_sign_subtree_request` doc comment carries forward a TODO from
the original `ietf_mtc_api::sign_subtree` module: the `\n\n` blank-line
separator between sections is ambiguous with the intra-note
text/signatures boundary. Two cleaner alternatives (length-prefixed
sections; TLS-binary framing) are sketched for upstream discussion.

Follow-ups:

  - `lvalenta/ietf-mtc-v2` drops the duplicate parsers +
    `serialize_mtc_subtree_signature_input` from `ietf_mtc_api`; its
    `MtcCosigner::sign_subtree` becomes a thin wrapper over
    `tlog_subtree_signature::sign_subtree`, and `MtcSubtreeNoteVerifier`
    becomes a thin wrapper that fixes the MTC identity convention in
    place around `SubtreeNoteVerifier`.
  - `lvalenta/tlog-witness` gains a `/sign-subtree` endpoint using this
    crate's parsers + `ietf_mtc_api` for the MTC signing semantics.

[C2SP#237]: https://github.com/C2SP/C2SP/pull/237

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2943,6 +2943,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "tlog_subtree_signature"
+version = "0.2.0"
+dependencies = [
+ "base64",
+ "byteorder",
+ "length_prefixed",
+ "signature",
+ "signed_note",
+ "thiserror 2.0.17",
+ "tlog_tiles",
+]
+
+[[package]]
 name = "tlog_tiles"
 version = "0.2.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
     "crates/signed_note",
     "crates/signed_note_wasm",
     "crates/static_ct_api",
+    "crates/tlog_subtree_signature",
     "crates/tlog_tiles",
     "crates/tlog_tiles_wasm",
     "crates/x509_util",
@@ -38,6 +39,7 @@ default-members = [
     "crates/signed_note",
     "crates/signed_note_wasm",
     "crates/static_ct_api",
+    "crates/tlog_subtree_signature",
     "crates/tlog_tiles",
     "crates/tlog_tiles_wasm",
     "crates/x509_util",
@@ -115,6 +117,7 @@ signature = "3.0.0-rc.10"
 signed_note = { path = "crates/signed_note", version = "0.2.0" }
 static_ct_api = { path = "crates/static_ct_api", version = "0.2.0" }
 thiserror = "2.0"
+tlog_subtree_signature = { path = "crates/tlog_subtree_signature", version = "0.2.0" }
 tlog_tiles = { path = "crates/tlog_tiles", version = "0.2.0" }
 tokio = { version = "1", features = ["sync"] }
 reqwest = { version = "0.12", features = ["json"] }
--- a/crates/tlog_subtree_signature/Cargo.toml
+++ b/crates/tlog_subtree_signature/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "tlog_subtree_signature"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+readme.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "Wire-format parsers, serializers, and the mtc-subtree/v1 binary signing format for the sign-subtree cosigner endpoint"
+categories = ["cryptography"]
+keywords = ["transparency", "subtree", "cosignature", "checkpoint"]
+
+[lib]
+crate-type = ["rlib"]
+
+[dependencies]
+base64.workspace = true
+byteorder.workspace = true
+length_prefixed.workspace = true
+signature.workspace = true
+signed_note.workspace = true
+thiserror.workspace = true
+tlog_tiles.workspace = true
--- a/crates/tlog_subtree_signature/src/crypto.rs
+++ b/crates/tlog_subtree_signature/src/crypto.rs
@@ -1,0 +1,398 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the BSD-3-Clause license found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+//! Binary signing format and signer/verifier abstraction for subtree
+//! cosignatures.
+//!
+//! The current binary message format is `mtc-subtree/v1\n\0`, defined in
+//! [draft-ietf-plants-merkle-tree-certs-02 §5.4.1][mtc-541]. This crate
+//! anticipates the format being renamed / restructured when the `sign-subtree`
+//! protocol migrates into its own C2SP specification (provisionally
+//! `c2sp.org/tlog-subtree-signature`), so the public helpers below are named
+//! without the `mtc-` prefix; the string `mtc-subtree/v1` appears only inside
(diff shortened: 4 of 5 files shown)


lukevalenta:
Thanks for the review! Addressed the feedback and added a second commit switching `bootstrap_mtc_api` onto the new crate (with a byte-pinned regression test).

**Commit 1** (`4e5ffe1`, autosquashed into the original): review follow-ups for `tlog_subtree_signature` itself.

- **(1) Forward-looking doc refs.** Replaced the `ietf_mtc_api::MtcSigningKey/...` references with prose pointing at `bootstrap_mtc_api` as the in-tree reference consumer and a note that the planned `ietf_mtc_api` will add ML-DSA later. Dropped the dead `[...]: https://docs.rs/ietf_mtc_api` footnotes from `lib.rs`, and rewrote the `crypto.rs` module-level reference similarly.
- **(3) Cross-reference on `parse_subtree_note_body_for_verify`.** Doc comment now explicitly names the richer-error sibling `parse_subtree_note_body` in `lib.rs` and the `parse_sign_subtree_request` caller path.
- **Nit: empty sigs in `serialize_sign_subtree_response`.** Added a `debug_assert!(!sigs.is_empty(), ...)` and split the test into a debug-build `#[should_panic]` and a release-build total-function case. Release behavior unchanged.
- **Nit: `MalformedRequest` carrying serialize-side errors.** Added `TlogSubtreeSignatureError::ConsistencyProofTooLarge(usize)` and routed both `parse_consistency_proof` (oversize line count) and `serialize_sign_subtree_request` (oversize slice) through it. `MalformedRequest` now only covers inbound-parse errors, as you suggested.
- **Nit: comment on BER bytes in pin test.** Added a multiline comment explaining the `\x82\xda\x4b\x30\x{01,02}` encoding (MTC `TrustAnchorID` per §4.1, Cloudflare-style experimental arc `1.3.6.1.4.1.44363.48.{1,2}`) matching what `bootstrap_mtc_api::RelativeOid` emits.
- **Tests:**
  - `parse_request_exact_max_proof_hashes`: 63-hash proof parses successfully (boundary coverage).
  - `parse_request_rejects_blank_trailing_proof_line`: pins the strict-trailing-blank-line behavior you flagged in (5).
  - `subtree_note_verifier_does_not_authenticate_origin`: same signature verifies against two different origin lines, pinning the "origin is not part of the signing input" contract.

On **(2)** (parser fragility around "each note has exactly one internal `\n\n`"): carrying this feedback upstream into the C2SP spec discussion; no change in this PR.

On **(4)** (silent-drop parse errors in `SubtreeNoteVerifier::verify`): left as-is since that's the correct `NoteVerifier::verify -> bool` contract, matching existing `MtcNoteVerifier` behavior.

**Commit 2** (`d80d71e`): **Wire `bootstrap_mtc_api` onto the `tlog_subtree_signature` crate.**

- `MtcCosigner::sign_subtree` delegates to `tlog_subtree_signature::sign_subtree` via a private `Ed25519RawSigner<'_>` newtype (orphan rules prevent impling the trait directly on `ed25519_dalek::SigningKey`). Analogous `Ed25519RawVerifier` in `MtcNoteVerifier::verify`.
- Deletes the local `serialize_mtc_subtree_signature_input` helper — single source of truth for the `mtc-subtree/v1\n\0` layout is now `tlog_subtree_signature::serialize_subtree_signature_input`.
- Adds a belt-and-suspenders binary-format regression test `mtc_subtree_v1_binary_format_unchanged` in `bootstrap_mtc_api::cosigner` that asserts the exact byte layout produced when plumbed end-to-end through `TrustAnchorID::from_str` + `RelativeOid` BER encoding + the shared crate's serializer. Duplicates the pin coverage in `tlog_subtree_signature::crypto` deliberately — if either side ever drifts, both tests should notice.
- No public-API changes to `bootstrap_mtc_api`.

All four pre-push checks clean locally. Also verified each commit on the branch compiles standalone (bisect-friendly).

lukevalenta:
Closing and will follow up with PRs of smaller scope that more closely track the latest spec developments.

PR #222, opened 2026-04-21
tlog_tiles: add evaluate_subtree_inclusion_proof helper

lukevalenta:
## Summary

Adds `tlog_tiles::evaluate_subtree_inclusion_proof`, the complement of the existing `verify_subtree_inclusion_proof`: given the proof and leaf hash, it derives the subtree root instead of checking an already-known root.

Needed by MTC-style cosignature verifiers that receive a subtree inclusion proof from the log and a cosignature over the subtree hash from the CA: they reconstruct the subtree hash from proof+leaf before handing it to the cosignature verifier. Today's `verify_*` variant is geared toward the opposite direction.

Implements draft-ietf-plants-merkle-tree-certs §4.3.2 "Evaluating a Subtree Inclusion Proof".

## Tests

- Extended the existing `test_tree` roundtrip (which already exercises every `(lo, hi, leaf_index)` combination for tree sizes 0..100) to also assert that the evaluator returns the expected subtree hash at every call site.
- New `test_evaluate_subtree_inclusion_proof_rejects_out_of_range` pins three failure modes (leaf below subtree, leaf at/above subtree, empty proof when one is required).

## Testing

`cargo clippy --workspace --all-targets -- -Dwarnings -Dclippy::pedantic`, `cargo test`, `cargo fmt --all --check`, `cargo machete` all pass locally.

### DIFF
--- a/crates/tlog_tiles/src/tlog.rs
+++ b/crates/tlog_tiles/src/tlog.rs
@@ -483,44 +483,16 @@ pub fn verify_inclusion_proof(
     leaf_index: u64,
     leaf_hash: Hash,
 ) -> Result<(), TlogError> {
-    // 1. Compare leaf_index from the inclusion_proof_v2 structure against tree_size. If leaf_index is greater than or equal to tree_size, then fail the proof verification.
-    if leaf_index >= tree_size {
-        return Err(TlogError::InvalidProof);
-    }
-    // 2. Set fn to leaf_index and sn to tree_size - 1.
-    let mut f_n = leaf_index;
-    let mut s_n = tree_size - 1;
-    // 3. Set r to hash.
-    let mut r = leaf_hash;
-    // 4. For each value p in the inclusion_path array:
-    for p in proof {
-        // a. If sn is 0, then stop the iteration and fail the proof verification.
-        if s_n == 0 {
-            return Err(TlogError::InvalidProof);
-        }
-        // b. If LSB(fn) is set, or if fn is equal to sn, then:
-        if lsb_set(f_n) || f_n == s_n {
-            // i. Set r to HASH(0x01 || p || r).
-            r = node_hash(*p, r);
-            // ii. If LSB(fn) is not set, then right-shift both fn and sn equally until either LSB(fn) is set or fn is 0.
-            //
-            // NOTE: It must be the case that fn is non-zero, so we can simplify.
-            while !lsb_set(f_n) {
-                f_n >>= 1;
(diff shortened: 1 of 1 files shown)


crates/tlog_tiles/src/tlog.rs
```diff
@@ -483,65 +483,116 @@ pub fn verify_inclusion_proof(
...
-    // 1. Compare leaf_index from the inclusion_proof_v2 structure against tree_size. If leaf_index is greater than or equal to tree_size, then fail the proof verification.
-    if leaf_index >= tree_size {
+    // Delegate to the evaluator with a synthetic `[0, tree_size)` subtree,
+    // matching how the MTC spec layers §4.3.3 on top of §4.3.2. A
+    // `tree_size` of 0 fails `Subtree::new` here (with `ConditionNotMet`
+    // rather than `InvalidProof`); this is a strict-subset behavioral
+    // change — nothing in the workspace matches on the specific error
+    // variant — and is the correct outcome, since an empty tree has no
+    // leaves to prove inclusion for.
+    let subtree = Subtree::new(0, tree_size)?;
```
lbaquerofierro, line 493: maybe a test for tree_size == 0 would be helpful here, in case someone later adds a tree_size == 0 shortcut and silently flips the error variant back to InvalidProof

crates/tlog_tiles/src/tlog.rs
```diff
@@ -1314,6 +1370,35 @@ mod tests {
...
+        assert!(evaluate_subtree_inclusion_proof(
+            &proof,
+            &Subtree::new(2, 4).unwrap(),
+            1,
+            leaves[1]
+        )
+        .is_err());
+        // leaf_index at or above subtree.hi is rejected.
+        assert!(evaluate_subtree_inclusion_proof(&proof, &subtree, 4, leaves[0]).is_err());
+        // Empty proof for a leaf index that requires one is rejected.
```
lbaquerofierro: tiny thing: the empty-proof case here actually fails on the final `s_n != 0` check rather than anything in the loop. might be worth tweaking the comment so that's a bit more obvious.

lukevalenta: Clarified in the comment 👍

crates/tlog_tiles/src/tlog.rs
lukevalenta: Good idea -- I added a test to pin the error variant.

Issue #208, opened 2026-04-13
Use native WebCrypto for ML-DSA


lukevalenta:
Switch to using native WebCrypto for ML-DSA support after https://github.com/cloudflare/workerd/pull/6403.

PR #227, opened 2026-04-30
Add subtree_v1 ML-DSA-44 cosignatures to tlog_cosignature

lukevalenta:
## Summary

Adds the `subtree/v1` cosignature format from [c2sp.org/tlog-cosignature](https://c2sp.org/tlog-cosignature): signed-note algorithm byte `0x06` paired with ML-DSA-44.

## Public API in `tlog_cosignature::subtree_v1`

- `build_cosigned_message`: algorithm-independent body builder for the `cosigned_message` TLS-presentation struct. Takes raw `(start, end)` rather than a typed `&Subtree` because consumers (notably future [draft-ietf-plants-merkle-tree-certs](https://datatracker.ietf.org/doc/draft-ietf-plants-merkle-tree-certs/) CA cosigners) may have already validated their inputs upstream and don't want a redundant construction; callers MUST ensure `[start, end)` is a valid subtree.
- `timestamped_signature`: `BE u64 timestamp || raw signature` blob used inside signed-note lines. Available for callers that produce signed-note-line output with a different signature algorithm.
- `SubtreeV1CheckpointSigner`: ML-DSA-44 signer producing `subtree/v1` cosignatures. Implements `tlog_tiles::CheckpointSigner` for the checkpoint case; `sign_subtree` covers arbitrary subtrees and takes `&tlog_tiles::Subtree` so subtree validity is enforced by the type system rather than at runtime. C2SP `tlog-cosignature` mandates ML-DSA-44 for `subtree/v1` signed-note lines, so this type is spec-faithful rather than generic.
- `SubtreeV1NoteVerifier`: matching ML-DSA-44 verifier. Symmetric shape: `verify_subtree` takes `&Subtree`; the `NoteVerifier::verify` path constructs `Subtree::new(0, checkpoint.size())` and bails on empty checkpoints.

## Spec invariants enforced at the API boundary

- `[start, end)` subtree validity (start < end, alignment to next-power-of-two width per draft-ietf-plants-merkle-tree-certs §4.1) is expressed in the type via `&Subtree`.
- `timestamp != 0` when `start != 0` is rejected on verify; `sign_subtree` panics on the same combination since it indicates caller error.
- `log_origin` is asserted to fit `opaque<1..2^8-1>`. The matching bound on `cosigner_name` is part of `KeyName`'s invariant (see below).
- `extract_timestamp_millis` rejects blobs whose length isn't exactly `TIMESTAMPED_SIGNATURE_LEN`, and uses `checked_mul` to guard seconds-to-millis conversion against overflow.

## `signed_note` changes

- Added `KeyName::MAX_LEN = 255` and enforce it in `KeyName::new`. The signed-note spec itself doesn't constrain length, but every C2SP consumer that embeds a key name in a length-prefixed wire format does so as `opaque<1..2^8-1>` — `cosigner_name` and `log_origin` in tlog-cosignature being the motivating example. Pushing the cap into the type means consumers can rely on the invariant without re-checking, and over-long names fail closed at construction. **Public-API tightening — call out in the next `signed_note` release notes.**
- Added `SignatureType::MlDsa44 = 0x06`.
- New `test_key_name_validation` covers empty / ASCII whitespace / Unicode whitespace / `+` / max-length acceptance / byte-vs-character distinction.

## Workspace plumbing

- `[workspace.dependencies]` adds `ml-dsa = "=0.1.0-rc.8"` (the `=` pin matches the wider RustCrypto pre-release pinning convention already documented in the comment block).
- `pkcs8` is tightened from `"0.11.0-rc.11"` to `"=0.11.0-rc.11"` so transitive resolvers don't pick the released `0.11.0` (whose API is incompatible with the rc.11 ed25519/ml-dsa pre-releases).
- `tlog_cosignature` gains `length_prefixed` (for the TLS-style length-prefixed `cosigner_name` / `log_origin` encoding) and `ml-dsa` dependencies.

## Test coverage

12 unit tests in `tlog_cosignature::subtree_v1`:
- Sign-then-verify roundtrips for both checkpoint (`start = 0`) and arbitrary-subtree (`start != 0`) cases.
- Spec's timestamp/start invariant.
- Mismatched-input rejection (different subtree, different hash, different origin).
- `NoteVerifier::verify` checkpoint reconstruction path through a real `signed_note::Note`.
- `extract_timestamp_millis` envelope-length and overflow handling.
- Key-ID derivation against the explicit `0x06` algorithm byte.
- Byte-pinned regression test on the `cosigned_message` layout (catches any drift in the TLS presentation).
- Pin on `Subtree::new`'s alignment rejection so the type-level validation can't quietly weaken.

## Followup

When the [IETF MTC draft-04](https://github.com/ietf-plants-wg/merkle-tree-certs/pull/215) lands, the MTC CA cosigner implementation (in a future `ietf_mtc_api` crate) will use `build_cosigned_message` directly with PKIX-defined signers (Ed25519, ECDSA, RSA, ML-DSA-44, ML-DSA-65, ML-DSA-87 per the draft). The signed-note-line wrapper isn't needed for cert-embedded cosignatures (the timestamp lives in the body); MTC sign-subtree implementations will use `timestamped_signature` directly with their own algorithm.

## Testing

`cargo clippy --workspace --all-targets -- -Dwarnings -Dclippy::pedantic`, `cargo test`, `cargo fmt --all --check`, `cargo machete` all pass locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,7 @@ jsonschema = "0.30"
 length_prefixed = { path = "crates/length_prefixed" }
 libfuzzer-sys = "0.4"
 log = { version = "0.4" }
+ml-dsa = "=0.1.0-rc.8"
 bootstrap_mtc_api = { version = "0.2.0", path = "crates/bootstrap_mtc_api" }
 p256 = { version = "0.14.0-rc.8", features = ["ecdsa"] }
 p384 = { version = "0.14.0-rc.8", features = ["ecdsa"] }
@@ -141,4 +142,4 @@ rsa = { version = "0.10.0-rc.17", default-features = false, features = ["sha2", 
 hashbrown = "0.15"
 spki = "0.8"
 const-oid = "0.10"
-pkcs8 = { version = "0.11.0-rc.11", features = ["pem"] }
+pkcs8 = { version = "=0.11.0-rc.11", features = ["pem"] }
--- a/crates/tlog_cosignature/src/subtree_v1.rs
+++ b/crates/tlog_cosignature/src/subtree_v1.rs
@@ -1,0 +1,804 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the BSD-3-Clause license found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+//! `subtree/v1` cosignatures, per [c2sp.org/tlog-cosignature][spec].
+//!
+//! Compared to the older Ed25519 [`cosignature/v1`][cosignature_v1] format
+//! (which is locked to checkpoints), `subtree/v1` is a cosignature that
+//! signs an arbitrary subtree of the log. A "checkpoint cosignature" is
+//! the special case `start = 0`, `end = checkpoint_size`.
+//!
+//! # What this module provides
+//!
+//! - [`build_cosigned_message`]: pure body builder for the
+//!   [`cosigned_message`][spec] struct. Algorithm-independent — used by
+//!   both the C2SP signed-note flow below and by other consumers (notably
+//!   [draft-ietf-plants-merkle-tree-certs][mtc], which embeds the same
+//!   body inside an X.509 certificate's `signatureValue`).
+//! - [`timestamped_signature`]: pure wrapper builder for the
+//!   `BE u64 timestamp || raw signature` blob used inside signed-note
+//!   lines. Available as a building block for callers that need to
+//!   produce signed-note lines but supply their own algorithm.
+//! - [`SubtreeV1CheckpointSigner`] / [`SubtreeV1NoteVerifier`]:
+//!   ML-DSA-44 signer / verifier for the C2SP-mandated signed-note
+//!   variant. C2SP `tlog-cosignature` mandates ML-DSA-44 for
+//!   `subtree/v1`; these types are spec-faithful and not generic.
+//!
+//! Consumers needing a different signature algorithm (e.g. an MTC CA
+//! cosigner using ECDSA or RSA per its own X.509 SPKI) should use
+//! [`build_cosigned_message`] (and [`timestamped_signature`] if they
+//! need signed-note-line output) directly with their own signer/verifier
+//! types; the algorithm-independent body builder is the abstraction
+//! point.
+//!
+//! # Signed message
+//!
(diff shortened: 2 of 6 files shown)


lukevalenta:
Ready for human review ✨ (I'll mark future PRs as draft until AI nits are cleaned up to hopefully reduce the noise)

crates/tlog_cosignature/src/subtree_v1.rs
```diff
@@ -0,0 +1,796 @@
...
+const LABEL: &[u8; 12] = b"subtree/v1\n\0";
+
+/// The encoded ML-DSA-44 signature length per FIPS 204.
+///
+/// `subtree/v1` currently has a single codepoint
+/// (`SignatureType::MlDsa44 = 0x06`). If `tlog-cosignature` allocates
+/// further codepoints, the parser/verifier helpers would generalize
+/// over a per-algorithm descriptor and this constant would be
+/// replaced.
+const MLDSA_44_SIGNATURE_LEN: usize = 2420;
```
bwesterb: Nit: can we easily pull this from the ml-dsa implementation?

Cargo.toml
```diff
@@ -104,6 +104,7 @@ jsonschema = "0.30"
 length_prefixed = { path = "crates/length_prefixed" }
 libfuzzer-sys = "0.4"
 log = { version = "0.4" }
+ml-dsa = "=0.1.0-rc.8"
```
bwesterb, line 107: We'll have to replace with Boring for FIPS in the future.

bwesterb: Had a quick look at the ml-dsa crate and it doesn't raise red flags. If we're going to use this in prod, we should do a proper review just to be sure.

bwesterb (approved):

Cargo.toml
lukevalenta: Also have https://github.com/cloudflare/azul/issues/208 to use the WebCrypto ML-DSA implementation once that's available.

crates/tlog_cosignature/src/subtree_v1.rs
lukevalenta: The obvious way doesn't work (`SignatureParams::SignatureSize` is `pub(crate)`), but found a different way!

PR #228, opened 2026-04-30
Add sign-subtree wire format to tlog_witness crate

lukevalenta:
## Summary

Adds wire-format parsers and serializers for the optional `sign-subtree` endpoint described in [c2sp.org/tlog-witness#sign-subtree](https://c2sp.org/tlog-witness#sign-subtree) (currently in draft at [C2SP PR #245](https://github.com/C2SP/C2SP/pull/245)). The endpoint lets a client ask a witness to cosign a subtree of a previously-cosigned tree, scoped to a `[start, end)` range, accompanied by a consistency proof to a reference checkpoint and (optionally) DoS-protection cosignatures from other known witnesses.

This PR is **wire-format only** — no crypto and no `witness_worker` handler. The handler that wires these parsers up to a `SubtreeV1CheckpointSigner` (PR #227) and switches the worker default to ML-DSA-44 will follow in a separate PR.

## Commits

1. **`signed_note: add Note::signatures() accessor`** — small symmetric addition next to `text()`. Lets callers count or inspect signatures without going through `Note::verify()` (which requires verifiers and may error). Needed by the sign-subtree serializer to decide whether to emit the checkpoint signature block and to enforce the spec's `MAX_CHECKPOINT_SIGNATURES = 8` cap.
2. **`tlog_witness: extract shared error and helpers into common module`** — pure refactor. Moves `TlogWitnessError`, `MAX_CONSISTENCY_PROOF_LINES`, `CONTENT_TYPE_TLOG_SIZE`, and `parse_proof_line` out of `add_checkpoint.rs` into a new private `common.rs`. No behavior change.
3. **`tlog_witness: add sign-subtree wire format`** — the new `sign_subtree` module with parsers, serializers, and 13 unit tests (request roundtrips, response roundtrip, spec-example shape pin, all bound and ordering rejection paths).

## Wire format covered

- **Request body**: `subtree <start> <end>\n` range line, base64 subtree hash, up to 8 subtree-cosignature lines (em-dash-prefixed), up to 63 consistency-proof hashes, blank-line separator, then a reference checkpoint that is either a full signed note or a bare body when no checkpoint signatures are supplied.
- **Response body**: identical shape to `add-checkpoint`'s response (one or more signed-note signature lines).

The parser enforces the spec's ordering rule (cosignature lines MUST appear before proof lines) and all spec-mandated upper bounds.

## What's NOT in this PR

- No `/sign-subtree` HTTP handler in `witness_worker` — that comes in a follow-up.
- No verification strategy (the spec lists three; the handler PR will hardcode the stateless one to start).
- No change to the witness's default cosignature algorithm — `add-checkpoint` still produces `cosignature/v1` Ed25519 signatures. The follow-up PR will add ML-DSA-44 / `subtree/v1` support and switch the default, deriving the algorithm from the OID in the configured PEM signing key.

## Pre-push checks

All four pass on each commit:

- `cargo test` (27 tlog_witness tests: 14 add_checkpoint + 13 sign_subtree, plus the rest of the workspace)
- `cargo fmt --all --check`
- `cargo machete`

### DIFF
--- a/crates/tlog_witness/src/sign_subtree.rs
+++ b/crates/tlog_witness/src/sign_subtree.rs
@@ -1,0 +1,654 @@
+// Copyright (c) 2025 Cloudflare, Inc.
+// Licensed under the BSD-3-Clause license found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+//! Parsers and serializers for the optional `sign-subtree` endpoint.
+//!
+//! Spec: <https://c2sp.org/tlog-witness#sign-subtree>.
+//!
+//! # Wire format
+//!
+//! The request body consists of:
+//!
+//! ```text
+//! subtree <start> <end>\n         ← subtree range (ASCII decimal, no leading zeros)
+//! <base64 subtree hash>\n         ← Merkle Tree hash of the subtree
+//! [— <name> <b64 cosig blob>\n]   ← 0..8 subtree cosignature lines (DoS protection)
+//! [<base64 proof hash>\n]         ← 0..63 consistency-proof hashes
+//! \n                              ← empty line separator
+//! <reference checkpoint>          ← either a signed-note checkpoint
+//!                                   (origin\nsize\nhash\n\n— sig\n...) or
+//!                                   a bare three-line checkpoint body
+//!                                   (origin\nsize\nhash\n) when no
+//!                                   checkpoint signatures are supplied.
+//! ```
+//!
+//! Per the spec, when checkpoint signatures are omitted the request body
+//! terminates with the newline that concludes the checkpoint root-hash
+//! line — no trailing blank line after the bare checkpoint body.
+//!
+//! Subtree cosignature lines are distinguished from consistency-proof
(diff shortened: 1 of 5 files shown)


crates/tlog_witness/src/sign_subtree.rs
```diff
@@ -0,0 +1,606 @@
...
+///
+/// # Errors
+///
+/// Returns [`TlogWitnessError::MalformedRequest`] if the request fails
+/// any structural check (missing required line, wrong line ordering,
+/// over-bound counts, malformed `subtree` line, malformed base64) or
+/// [`TlogWitnessError::Note`] if any cosignature or checkpoint-signature
+/// line fails [`NoteSignature::from_bytes`] / the embedded note fails
+/// [`Note::from_bytes`].
+pub fn parse_sign_subtree_request(body: &[u8]) -> Result<SignSubtreeRequest, TlogWitnessError> {
```
lukevalenta, line 114: Should we set a limit on the size we're willing to parse?

rozbb (approved):

PR #236, opened 2026-05-05
Add tlog_mirror crate with wire format and default ticket authenticator

lukevalenta:
NOTE(lvalenta): I used Claude to help generate this PR, but have reviewed and approve the changes.

Introduces the `tlog_mirror` crate, implementing the wire-format pieces of [c2sp.org/tlog-mirror](https://c2sp.org/tlog-mirror).


Cargo.toml
```diff
@@ -103,6 +109,7 @@ futures-executor = "0.3.31"
 futures-util = "0.3.31"
 getrandom = { version = "0.4", features = ["wasm_js"] }
 hex = "0.4"
+hmac = "0.13"
```
lbaquerofierro: this might be a leftover, nothing is pulling hmac

crates/tlog_mirror/src/ticket.rs
```diff
@@ -0,0 +1,181 @@
...
+    Aes256GcmSiv, Nonce,
+    aead::{Aead, KeyInit, Payload},
+};
+
+use crate::error::TicketError;
+
+/// Length of the AES-GCM-SIV authentication tag, in bytes.
+pub const TAG_LEN: usize = 16;
+
+/// Fixed nonce; deterministic sealing is safe under AES-GCM-SIV's
```
lbaquerofierro: RFC 8452 section 9 explicitly recommends against fixing the nonce. This reuses one nonce for every ticket under a potentially long-lived key without any usage bound. Would it be possible to generate a random nonce per ticket and encode nonce || ciphertext || tag, or use a construction intended for deterministic authentication? (or remove the claim is safe)

crates/tlog_mirror/src/wire.rs
```diff
@@ -0,0 +1,651 @@
...
+}
+
+impl EntryPackage {
+    /// Read one entry package. `num_entries` must be `end - start` for the
+    /// package's range from [`package_ranges`].
+    ///
+    /// # Errors
+    /// [`ParseError::Io`] on short reads, or [`ParseError::TooManyHashes`]
+    /// if `num_hashes` exceeds 63.
+    pub fn read_from<R: Read>(mut reader: R, num_entries: u64) -> Result<Self, ParseError> {
```
lbaquerofierro, line 207: The specification limits each entry package to at most 256 entries, but read_from accepts any u64 and passes it to Vec::with_capacity. An incorrect count could cause a very large allocation or panic. Could we reject counts above PACKAGE_ALIGNMENT before allocating? It may also be worth rejecting packages with more than 256 entries in write_to, so the type cannot serialize an invalid package

crates/tlog_mirror/src/wire.rs
```diff
@@ -0,0 +1,651 @@
...
+        })
+    }
+
+    /// Write the header.
+    ///
+    /// # Errors
+    /// [`io::ErrorKind::InvalidInput`] if `log_origin` or `ticket` exceeds
+    /// the u16 length-prefix limit (65535 bytes); otherwise the writer's
+    /// IO errors.
+    pub fn write_to<W: Write>(&self, mut writer: W) -> io::Result<()> {
```
lbaquerofierro, line 104: read_from already rejects upload_start > upload_end, but it looks like write_to can still serialize that range. Since the draft requires upload_start <= upload_end, could we add the same validation here before writing anything, along with a serialization test? to keep the read and write paths consistent

crates/tlog_mirror/src/wire.rs
```diff
@@ -0,0 +1,651 @@
...
+/// whitespace, and any non-ASCII-decimal byte, enforcing one canonical
+/// encoding per integer.
+fn parse_decimal_u64(field: &'static str, value: &str) -> Result<u64, ParseError> {
+    if value.is_empty() {
+        return Err(ParseError::InvalidDecimal {
+            field,
+            value: value.to_owned(),
+        });
+    }
+    if value.len() > 1 && value.starts_with('0') {
```
lbaquerofierro: I don't think the draft requires canonical decimal formatting here, it only says these fields are decimal. That means something like 00042 should still be valid, and this check could make us reject responses that Tessera accepts. We can drop the leading-zero check and accept any non-empty digit string that fits in a u64? We can still keep to_body canonical when serializing

Cargo.toml
lukevalenta: Oops, good catch. Looks like cargo machete doesn't handle workspace dependencies.. I'll switch to cargo-shear in a followup PR.

crates/tlog_mirror/src/ticket.rs
lukevalenta: I initially used a fixed nonce since all we can about is authentication and not confidentiality for these tickets, but I'll switch to a random nonce just in case this code gets copied later!

crates/tlog_mirror/src/wire.rs
lukevalenta: Probably won't hit this case in practice, but seems fine to match what Tessera allows 👍

lbaquerofierro (approved):

PR #242, opened 2026-06-17
Migrate to axum::Router

mendess:
The `axum::Router` is much more flexible than the native `worker::Router`
implementation and makes use of the rust http ecosystem. It also allows for type
safe request parameter validation as well as middleware. Which is usefull when
trying to instrument the workers for production (adding metrics, etc). Which is
something I'll be adding in a follow up PR.

- **Factor out route handlers to seperate functions**
    First the request handling logic was factored out to seperate functions
- **Migrate to axum::Router**
    And after that they were migrated to use axum, in the hope that this helps
    review the code.

There is some code duplication in this PR, but it's the "same amount" of code duplication that was already present in the codebase. The deduplication work is left as a future exercise to avoid overloading this PR.

### DIFF
--- a/crates/bootstrap_mtc_worker/src/frontend_worker.rs
+++ b/crates/bootstrap_mtc_worker/src/frontend_worker.rs
@@ -107,156 +107,25 @@ async fn main(req: Request, env: Env, ctx: Context) -> Result<Response> {
             // handle the request.
             Router::with_data(name)
                 .get_async("/logs/:log/get-roots", |_req, ctx| async move {
-                    Response::from_json(&GetRootsResponse {
-                        certificates: x509_util::certs_to_bytes(
-                            &load_roots(&ctx.env, ctx.data).await?.certs,
-                        )
-                        .unwrap(),
-                    })
+                    get_roots(&ctx.env, ctx.data).await
                 })
                 .post_async("/logs/:log/add-entry", |req, ctx| async move {
                     add_entry(req, &ctx.env, ctx.data).await
                 })
-                .post_async("/logs/:log/get-certificate", |mut req, ctx| async move {
-                    let name = ctx.data;
-                    let params = &CONFIG.logs[name];
-                    let Ok(GetCertificateRequest {
-                        leaf_index,
-                        spki_der,
-                    }) = req.json().await
-                    else {
-                        return Response::error("Unexpected input", 400);
-                    };
-                    let object_backend = ObjectBucket::new(load_public_bucket(&ctx.env, name)?);
-                    // Fetch the current checkpoint to know which tiles to fetch
-                    // (full or partials).
-                    let (checkpoint, _checkpoint_bytes) =
-                        get_current_checkpoint(&ctx.env, name, &object_backend).await?;
-                    if leaf_index >= checkpoint.size() {
-                        return Response::error("Leaf index is not in log", 422);
-                    }
-
-                    let seq = get_landmark_sequence(name, &object_backend).await?;
-                    if leaf_index < seq.first_index() {
-                        return Response::error("Leaf index is before first active landmark", 422);
-                    }
-                    let Some((landmark_id, landmark_subtree)) = seq.subtree_for_index(leaf_index)
-                    else {
-                        // The leaf index might be between the latest landmark
-                        // and the current tree size. Set Retry-After to the
-                        // expected time for the next landmark so the client can
-                        // try again later.
-                        let headers = Headers::new();
-                        let i = params.landmark_interval_secs as u64;
-                        headers
-                            .set("Retry-After", &format!("{}", i - (now_millis() / 1000) % i))?;
-                        return Response::error("Leaf index will be covered by next landmark", 503)
-                            .map(|r| r.with_headers(headers));
-                    };
-
-                    // Fetch the log entry for the leaf index.
-                    let log_entry = read_leaf::<BootstrapMtcLogEntry>(
-                        &object_backend,
-                        leaf_index,
-                        checkpoint.size(),
-                        checkpoint.hash(),
-                    )
-                    .await
-                    .map_err(|e| e.to_string())?;
-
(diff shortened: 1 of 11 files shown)


crates/bootstrap_mtc_worker/src/frontend_worker.rs
```diff
@@ -82,277 +90,180 @@ fn start() {
...
+    NotFound,
+    BadRequest(String),
+    UnknownLog,
+    FailedToSerializeSignaturelessCert(bootstrap_mtc_api::MtcError),
+    SubtreeInclusionProofFailed(tlog_core::TlogError),
+    LeafIndexBeforeFirstActiveLandmark,
+    LeafIndexNotInLog,
+    RedirectToMonitorApi(&'static str),
+    // TODO: needs a better name
+    LeafIndexCantBeCoveredYet { retry_after: u64 },
```
mendess: This is a question for reviewers. I'm not sure this name is very good

crates/bootstrap_mtc_worker/src/frontend_worker.rs
```diff
@@ -449,28 +365,258 @@ async fn add_entry(mut req: Request, env: &Env, name: &str) -> Result<Response>
...
+                axum::body::Body::from_stream(WorkerByteStream::new(body)),
+            ))
+        } else {
+            Err(AppError::NotFound)
+        }
+    } else {
+        // TODO: should this be an HTTP redirect instead of a 404?
+        Err(AppError::RedirectToMonitorApi(
+            &CONFIG.logs[&log].monitoring_url,
+        ))
```
mendess, line 556: This is also a question for reviewers, why not an HTTP 301?

crates/bootstrap_mtc_worker/src/frontend_worker.rs
```diff
@@ -82,277 +90,180 @@ fn start() {
...
-        });
+        .route("/logs/{log}/get-roots", get(get_roots))
+        .route("/logs/{log}/add-entry", post(add_entry))
+        .route("/logs/{log}/get-certificate", post(get_certificate))
+        .route("/logs/{log}/get-landmark-bundle", get(get_landmark_bundle))
+        .route("/logs/{log}/metadata", get(metadata))
+        .route("/logs/{log}/sequencer_id", get(sequencer_id))
+        .route("/logs/{log}/{*key}", get(get_object))
+        // The previous native router applied no body-size limit (bodies are
+        // bounded by the Workers runtime), so disable axum's 2 MiB default.
```
lukevalenta: A 2 MiB limit on request body size would be fine here. `add-entry` requests will be only a few KiBs max.

crates/bootstrap_mtc_worker/src/frontend_worker.rs
lukevalenta: Maybe this?
```suggestion
    LeafIndexPendingLandmark { retry_after: u64 },
```

crates/bootstrap_mtc_worker/src/frontend_worker.rs
lukevalenta: For production static CT logs, we'll want clients to hit the monitoring API (R2 bucket) directly, since it's cheaper and more cacheable, so I think it's better to give a hard failure if clients try to fetch tiles from the issuance API (this worker).

I think I only added the ability to leave 'monitoring_url' empty since it was convenient for testing. I don't remember if the current integration tests require it, but we could remove the "/logs/{log}/{*key}" route entirely otherwise to remove code bloat.

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -56,156 +64,272 @@ fn start() {
...
-                log::warn!("Internal error: {e}");
-                Response::error("Internal error", 500)
-            }
-        });
+        .route("/logs/{log}/ct/v1/get-roots", get(get_roots))
+        .route("/logs/{log}/ct/v1/add-chain", post(add_chain))
+        .route("/logs/{log}/ct/v1/add-pre-chain", post(add_pre_chain))
+        .route("/logs/{log}/log.v3.json", get(log_v3_json))
+        .route("/logs/{log}/sequencer_id", get(sequencer_id))
+        .route("/logs/{log}/{*key}", get(get_object))
```
lukevalenta, line 82: Same comment applies as for the bootstrap_mtc_worker: we can drop this route unless we want to keep it around for testing purposes. If we do drop it, we should update the 'config' crates to require 'monitoring_url' to be specified.

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -56,156 +64,272 @@ fn start() {
...
+        // The previous native router applied no body-size limit (bodies are
+        // bounded by the Workers runtime), so disable axum's 2 MiB default.
```
lukevalenta: Same comment as for bootstrap_mtc_worker: a default 2 MiB limit would be fine.

crates/ct_worker/src/frontend_worker.rs
```diff
@@ -56,156 +64,272 @@ fn start() {
...
+            mmd: MAX_MERGE_DELAY_SECS,
+            temporal_interval: &params.temporal_interval,
+        })
+        .unwrap(),
+    )
+        .into_response())
+}
+
+/// `GET /logs/{log}/sequencer_id`
+#[worker::send]
```
lukevalenta, line 263: This (as well as some of the other boilerplate) could probably be moved to `generic_log_worker` as it's duplicated across boostrap_mtc_worker and ct_worker, but OK to punt on that until later.

lukevalenta (approved):

crates/bootstrap_mtc_worker/src/frontend_worker.rs
mendess: Makes sense. I'll keep that in mind and consider removing the route at a later date

PR #244, opened 2026-06-23
Make roots cache per shard instead of global for all shards

mendess:


crates/ct_worker/src/lib.rs
```diff
@@ -23,6 +23,7 @@ mod sequence_metadata;
 mod sequencer_do;
 
 pub(crate) use sequence_metadata::StaticCTSequenceMetadata;
+use worker::send::SendWrapper;
```
lukevalenta: Maybe add to the existing worker imports above?

crates/ct_worker/src/lib.rs
```diff
@@ -100,26 +100,62 @@ pub(crate) fn load_origin(name: &str) -> KeyName {
...
-    if let Some(pool) = ROOTS.get() {
-        return Ok(pool);
+struct CachedRoot {
+    not_after: SendWrapper<Date>,
+    pool: Arc<CertPool>,
+}
+
+impl CachedRoot {
+    fn new(pool: CertPool) -> Self {
+        // this is an extreme case where a worker isolate lasts for more than a day, which is the
```
lukevalenta, line 109: If it cuts down on complexity, I think it would be fine to make each shard's roots a `&'static CertPool`. If a worker isolate lasts a long time and runs with a stale copy of the roots, the only impact is that it might reject some `add-chain`/`add-pre-chain` requests, which isn't a big deal.

That said, the current version looks good!

lukevalenta (approved):

crates/ct_worker/src/lib.rs
mendess: The Mutex doesn't allow us to return static references to objects guarded by it, since mutation implies that object could be dropped and having a cache that never expires feels wrong.

PR #250, opened 2026-07-07
Add sentry support

mendess:
Add sentry support to all workers. This requires a bump of the worker version to
support unwinding in wasm

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -47,7 +47,7 @@ jobs:
               exit 0
             fi
             echo "Waiting for wrangler dev... attempt $i/30"
-            sleep 2
+            sleep 8
           done
           echo "wrangler dev failed to start in time"
           exit 1
@@ -96,7 +96,7 @@ jobs:
               exit 0
             fi
             echo "Waiting for wrangler dev... attempt $i/30"
-            sleep 2
+            sleep 8
           done
           echo "wrangler dev failed to start in time"
           exit 1
@@ -144,7 +144,7 @@ jobs:
               exit 0
             fi
             echo "Waiting for wrangler dev... attempt $i/30"
-            sleep 2
+            sleep 8
           done
           echo "wrangler dev failed to start in time"
           exit 1
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -50,7 +50,7 @@ jobs:
               exit 0
             fi
             echo "Waiting for wrangler dev... attempt $i/30"
-            sleep 8
+            sleep 2
           done
           echo "wrangler dev failed to start in time"
           exit 1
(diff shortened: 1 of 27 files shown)


crates/generic_log_worker/src/obs/sentry.rs
```diff
@@ -0,0 +1,339 @@
...
+    let dsn = &transport.dsn;
+    let url = format!(
+        "{}://{}:{}/api/{}/envelope/",
+        dsn.scheme(),
+        dsn.host(),
+        dsn.port(),
+        dsn.project_id(),
+    );
+    let auth = format!(
+        "Sentry sentry_key={}, sentry_version=7, sentry_client=azul/{}",
```
lukevalenta, line 354: Aside from `sentry_client=azul` line, it seems like this sentry functionality would be pretty generally useful for Rust-based Workers. Worth considering publishing as a standalone `worker-sentry` crate, possibly in the https://github.com/cloudflare/workers-rs repo.

lukevalenta:
Since Bonk seems to be having trouble lately, here's some feedback from a local AI review:

Review summary of PR #250
CI: All three integration jobs are currently failing — not a compile error, but wrangler dev failed to start in time. The wasm build got slower (--panic-unwind, dropping strip = true, worker 0.7→0.8, plus sentry/getrandom deps) and likely blew past the ~60s startup window. Needs confirming whether to bump the timeout or if the build actually hangs.
Bugs / issues found:
1. Uuid::nil() on all panic events (Medium) — Sentry dedupes by event_id, so every panic sharing the all-zero ID may be dropped after the first. Since panic capture is the headline feature, this undermines it. Fix: derive a getrandom-free varying UUID (hash of message + file:line:col).
2. Cron handlers don't catch panics (Low–Medium) — the scheduled handlers capture Err but aren't wrapped in catch_unwind_and_flush, so a panic buffers an event that never flushes. Inconsistent with the DO/frontend paths.
3. witness_worker only partially integrated (Low) — gets --panic-unwind + RefUnwindSafe but no init_sentry/capture wrappers, despite "all workers" in the description.
Non-blocking: scope tags persist across requests in a DO isolate (use with_scope); DO errors reported at Fatal (defensible); trailing flush skipped on frontend ? error path (matches pre-existing behavior).

.github/workflows/integration.yml
```diff
@@ -26,11 +26,14 @@ jobs:
         run: cargo install wasm-pack@0.13.1 --locked
 
       - name: Install worker-build
-        run: cargo install worker-build@0.7.5 --locked
+        run: cargo install worker-build@0.8.5 --locked
 
+      # RUSTFLAGS: force legacy Wasm EH so wasm-bindgen 0.2.126's wasmparser
+      # accepts the module (modern exnref opcodes from nightly >=2026-05-09
+      # are not yet supported).
```
lukevalenta, line 33: From the commit message, looks like https://github.com/wasm-bindgen/wasm-bindgen/pull/5219 tracks support being added to wasm-bindgen without the flag? Could you link that here?

lukevalenta (approved):
LGTM. Use of nightly features seems OK.

PR #260, opened 2026-07-17
mirror_worker: scaffold the crate + config

lukevalenta:
NOTE(lvalenta): I used AI to help generate this PR, but I've reviewed and approve these changes. The mirror is targeted for an MTC deployment, so we exclude general tlog-mirror support for things like Ed25519 signatures, and bake log numbers into the config.

Add a deployable skeleton for a c2sp.org/tlog-mirror mirror, specialized
for MTC issuance logs. This wires the crate into the workspace and
validates the embedded per-environment config; the submission API,
cosigner keys, per-origin Durable Objects, and read interface follow in
later slices.


mendess (approved):

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -0,0 +1,39 @@
...
+async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
+    Router::new()
+        .get("/", |_req, _ctx| {
+            Response::ok(format!(
+                "{} - c2sp.org/tlog-mirror mirror\n",
+                CONFIG.mirror_name
+            ))
+        })
+        .run(req, env)
+        .await
```
mendess: use axum::Router instead

mendess (changes requested):

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Good catch. I'll also add this to AGENTS.md.

lukevalenta: Done!

crates/mirror_worker/config/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+    ///    and the window spans at most [`MAX_LOG_NUMBER_WINDOW`] log
+    ///    numbers.
+    /// 6. The longest origin from [`LogParams::origins`] still fits the
+    ///    signed-note key name length cap, since each origin is itself
+    ///    used as a checkpoint origin.
+    ///
+    /// `log_key_name` uniqueness across log entries is *not* checked here
+    /// because it is already enforced by the JSON object shape: duplicate
+    /// keys in `logs` are duplicate JSON object keys, which `serde_json`
+    /// rejects at deserialization time.
```
lbaquerofierro: `serde_json` doesn't actually reject duplicate object keys, it keeps the last value silently. The JSON-schema `propertyNames` check only validates key format, not uniqueness. So a duplicated `log_key_name` in `logs` wouldn't error, the last one just wins. Not necessarily worth adding a real dedup check (since the config is added manually). Worth just correcting the comment.

lukevalenta: Ah, good catch. Adding a check for duplicated keys isn't too bad with a custom deserializer, so I'll just do that to catch any future copy/paste bugs.

mendess: TIL

lbaquerofierro (approved):

mendess (approved):

PR #261, opened 2026-07-20
mirror_worker: cosigner key, /metadata, and add-checkpoint

lukevalenta:
Build on the crate skeleton (#260) with the mirror's key handling and the pending-checkpoint submission path:

- Parse trusted per-origin log keys (ML-DSA-44 / subtree/v1) from config and verify submitted checkpoints against them.
- Load the mirror's own ML-DSA-44 cosigner key and publish its SPKI, algorithm, and per-log config at GET /metadata.
- Handle POST /add-checkpoint via a per-origin MirrorState Durable Object, whose single-threaded fetch handler makes the check-verify-persist update of the pending checkpoint atomic.

The MirrorState DO also defines committed and next-entry state and RPCs beyond what add-checkpoint uses; these are covered by the DO's format/round-trip unit tests and will be used by the add-entries endpoint.


lukevalenta:
>    * **Unused import suppression:** The line `use ml_dsa::pkcs8::{DecodePrivateKey as _, EncodePublicKey as _};` imports `DecodePrivateKey` only for the underscore alias, which is fine but could be simplified to just `EncodePublicKey` if the decode path isn’t needed elsewhere. This is purely stylistic and does not affect functionality.

`DecodePrivateKey` is needed for `from_pkcs8_pem`

>    * **Error message consistency:** In `dispatch_update_pending`, the generic `Bad request` prefix is used for 400 responses from the DO. Consider mirroring the exact error text from the DO for easier debugging (e.g., forward the DO’s body verbatim). This is optional.

Addressed.

>    * **Header handling in `tlog_size_conflict`:** The `Headers` object is built and mutated before being attached to the response; this works, but using `Response::from_body(...).with_status(409).with_headers(headers)` could be expressed as a single chained call for readability.

Already in the single-chained form.

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -37,3 +65,264 @@ async fn fetch(req: Request, env: Env, _ctx: Context) -> Result<Response> {
...
+                    h
+                },
+                ..Default::default()
+            },
+        )?)
+        .await?;
+    match resp.status_code() {
+        200 => {
+            // Drain the body so we can drop the response.
+            let _ = resp.bytes().await?;
```
mendess: Is this really necessary? I know you have to do body.close() in JS but in rust the destructor should take care of that for you. Doing it this way slows down the worker

mendess (dismissed):

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Oh yep you're right. Thanks!

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -0,0 +1,328 @@
...
+/// atomically; see [`dispatch_update_pending`] for the status-code
+/// mapping.
+///
+/// [add-cp]: https://c2sp.org/tlog-mirror#add-checkpoint
+/// [`MirrorState`]: crate::mirror_state_do
+async fn add_checkpoint(mut req: Request, env: Env) -> Result<Response> {
+    // Cap the body so a misbehaving client can't make us buffer
+    // arbitrary data; anything larger is rejected downstream anyway.
+    let body = req.bytes().await?;
+    if body.len() > MAX_ADD_CHECKPOINT_BODY_SIZE {
```
lbaquerofierro: small thing: `req.bytes()` buffers the whole body first and *then* we check the length, so the cap doesn't actually stop us from buffering a big payload, it just rejects after we've already read it into memory. if you want the cap to mean something memory-wise, you could peek at the `content-length` header and bail early before reading the body. not a huge deal since the platform bounds body size anyway, just flagging it since the comment says we won't buffer arbitrary data.

lukevalenta: Fixed!

lukevalenta:
Latest push cleans up some minor docs issues.

lbaquerofierro (approved):

PR #264, opened 2026-07-24
mirror_worker: implement the add-entries submission API

lukevalenta:
NOTE(lvalenta): I separated https://github.com/cloudflare/azul/pull/267 (streaming in the add-entries request body) out from this commit to try to make it slightly easier to review. Integration tests are also deferred to a later PR.

Read and verify entry packages against the target pending checkpoint,
persist them as tlog-tiles entry bundles and hash tiles (see commit.rs),
and advance the persisted-entry frontier. A complete upload cosigns the
mirror checkpoint and returns 200; a client-truncated upload persists the
verified prefix and returns 202 with the advanced next entry so the
client can resume (C2SP/C2SP#253).

The whole request body is buffered and all its packages are committed in
a single pass. commit.rs persists resumably from the current frontier, so
a client that truncated at a package boundary resumes cleanly on its next
request. A follow-up commit adds incremental streaming so a large upload
does not have to buffer the entire body.

Includes body-decoding (transparent gzip), per-origin prefixed R2
storage, and the ticket-sealer for recovering a past pending checkpoint
on resume.


lukevalenta:
/bonk review again, hopefully this time without caching

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,898 @@
...
+        .map_err(|e| Error::from(format!("mirror cosign: {e:?}")))?;
+    let cosig_body =
+        tlog_witness::serialize_add_checkpoint_response(std::slice::from_ref(&note_sig));
+
+    // Persist the served checkpoint (log note + mirror cosignature) before
+    // advancing the durable mirror checkpoint, so a monitor can never
+    // observe an advanced size without the matching checkpoint object.
+    let mut checkpoint_obj = target.signed_note_bytes.clone();
+    checkpoint_obj.extend_from_slice(&cosig_body);
+    commit::write_checkpoint(bucket, checkpoint_obj.clone()).await?;
```
lbaquerofierro: This ordering protects the DO but I think it lets R2 go backwards. Two concurrent requests, both read committed = 100:

- B targets 500, lands first: R2 = 500, DO = 500
- A targets 200, lands second: R2 = 200, DO no-ops (size < committed.size)

Monitors read R2, so they'd see 500 then 200. The MUST NOT rewind guarantee only holds on the DO side.

Can we make the R2 write conditional on the commit actually advancing? The DO already knows, so if /commit returned whether it wrote we'd only upload on a real advance. Puts us back to commit-then-write, but the DO response tells us it's safe by then.

crates/mirror_worker/src/body.rs
```diff
@@ -0,0 +1,127 @@
...
+    while let Some(chunk) = stream.next().await {
+        let chunk = chunk.map_err(|e| Error::from(e.to_string()))?;
+        raw.extend_from_slice(&chunk);
+    }
+
+    match encoding.as_str() {
+        "" | "identity" => Ok(raw),
+        "gzip" | "x-gzip" => {
+            let mut out = Vec::new();
+            GzDecoder::new(raw.as_slice())
```
lbaquerofierro, line 63: Should we cap the decompressed size? out is unbounded. The 100 MB edge limit only bounds the compressed body, and deflate hits ~1032:1, so a 100 KB request can inflate to 100 MB and a large one blows past the 128 MB isolate ceiling you mention in the module doc

crates/mirror_worker/src/commit.rs
```diff
@@ -0,0 +1,466 @@
...
+        read_edge_tiles(object, persisted_size, &persisted_hash).await?
+    };
+
+    // Load the current partial entry bundle so we extend rather than
+    // overwrite it. Only exists when the frontier is mid-tile.
+    let mut data_tile = Vec::new();
+    if persisted_size > 0 && !persisted_size.is_multiple_of(TILE_WIDTH) {
+        let partial = TlogTile::from_index(stored_hash_index(0, persisted_size - 1))
+            .with_data_path(PathElem::Entries);
+        data_tile = object.fetch(partial.path()).await?.ok_or_else(|| {
```
lbaquerofierro, line 329: This trusts the fetched partial bundle without checking it. The new leaves get hashed from edge_tiles, which is authenticated against `persisted_hash` but these reloaded bytes go straight back out as the served entry bundle. So if the object were ever short or stale we'd publish a bundle whose leaves don't match the hash tiles, and the recomputed rot check at the end of persist_entries would still pass because it never looks at these bytes.
Maybe verify after the fetch that it decodes to exactly persisted_size % TILE_WIDTH length-prefixed entries with nothing left over?

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,898 @@
...
+            );
+            return Ok(mirror_info_409(&env, &snapshot, &header.log_origin));
+        }
+    };
+
+    // upload_start must be <= next_entry (the first index not yet
+    // persisted); a client resuming after a 202 sets it to the advertised
+    // next entry. A non-256-aligned value is accepted; see
+    // `first_package_prefix`.
+    if header.upload_start > snapshot.next_entry.size {
```
lbaquerofierro, line 143: This covers the upload_start too high case, but I don't see a bound on the too-low one. Spec has a 409 for when excess_entries = min(upload_end, next_entry) - upload_start is "too large" (threshold left to us) (without it a client can send upload_start = 0 against a large `next_entry` and make us re-parse and reverify the whole prefix, full Merkle recompute per package plus an R2 read in `first_package_prefix`)

crates/mirror_worker/src/body.rs
lukevalenta: Ah yeah, this'll be addressed in https://github.com/cloudflare/azul/pull/267 with the streaming approach, since we unzip and parse the request body incrementally and have a cap on memory consumption. Not a problem if the decompressed payload is over 100MB since we'll validate and commit the entry packages as we go.

crates/mirror_worker/src/add_entries.rs
lukevalenta: Oh yep, good catch. Will push a fix shortly to add the excess_entries bound. Using one package size to start, but we can tune it to be more forgiving later if needed.

lukevalenta: Addressed with https://github.com/cloudflare/azul/pull/264/commits/69693a8f1273cb5abb892ab33ee57ac7fbe7ed83

crates/mirror_worker/src/commit.rs
lukevalenta: Addressed with https://github.com/cloudflare/azul/pull/264/commits/e2991b3bb5c65f2db385a87b2ba40b590313273e

crates/mirror_worker/src/add_entries.rs
lukevalenta: Oof, yeah that was sloppy. Addressed with https://github.com/cloudflare/azul/pull/264/commits/fc238ed31d441f11f405f1ee0407d2fcfb8d2135 to move the R2 write into the DO, and wrap both the DO commit and R2 write with a Mutex guard (with DO commit first).

lukevalenta:
> #### 4. `CommitRequest.signed_note_bytes` carries the cosigned note, not the log's original
> `add_entries.rs:410-421` – `cosign_and_serve` constructs the served checkpoint as `target.signed_note_bytes + cosig_body` and sends that to the DO via `CommitRequest.signed_note_bytes`. The DO stores this in `CommittedCheckpoint.signed_note_bytes` (line 336). The doc comment on `CommittedCheckpoint.signed_note_bytes` (`mirror_state_do.rs:79`) says "Full signed-note bytes as the log signed them" — but after this PR the value stored there is actually the log's note _plus_ the mirror's cosignature lines. The comment should be updated to match reality.

Fixed with https://github.com/cloudflare/azul/pull/264/commits/90e540322d9232f0e204c0202be91aabdb29bd3e. None of the other comments require changes.

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,976 @@
...
+/// Returns `Err(reason)` (a static `&str` describing why) when
+/// neither path produces a target. The frontend turns the error into
+/// a 409 with `text/x.tlog.mirror-info`.
+fn resolve_target_pending(
+    env: &Env,
+    header: &AddEntriesRequestHeader,
+    snapshot: &MirrorStateSnapshot,
+    verifiers: &signed_note::VerifierList,
+) -> std::result::Result<PendingCheckpoint, &'static str> {
+    if header.upload_end == snapshot.pending.size {
```
lbaquerofierro, line 675: I think this misses one of the two values the spec says we MUST accept for upload_end. It's "neither a known pending checkpoint value nor the mirror checkpoint's tree size", and we only cover current-pending plus whatever the ticket carries

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,976 @@
...
+            env,
+            snapshot,
+            &header.log_origin,
+            frontier_size,
+        ));
+    }
+
+    // Complete upload: no trailing bytes may remain after the last package.
+    let pos = usize::try_from(cursor.position()).unwrap_or(usize::MAX);
+    if pos < cursor.get_ref().len() {
```
lbaquerofierro: Spec says to discard these bytes rather than reject: "The mirror discards any partial bytes after the last successfully authenticated entry package." The only 400 it defines is the case where no package was authenticated at all.

The ordering is a bigger issue though. `persist_entries` and `advance_next_entry` have already run and awaited by this point, so one stray trailing byte means we durably write every tile, move the DO fron, and then return a 400 with no mirror-info body. The client reads that as "nothing was saved" when in fact everything was, and gets no `next_entry` to resume from. Any client with an off-by-one on its last package length hits this every time.

Maybe just drop the `check and log::warn!?` If you'd rather keep it, it can be moved above the persist call to make the 400 honest

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,976 @@
...
+        &header.log_origin,
+        &CommitRequest {
+            size: header.upload_end,
+            hash: target.hash,
+            signed_note_bytes: checkpoint_obj,
+        },
+    )
+    .await?;
+
+    Ok((
```
lbaquerofierro, line 512: I think a skipped checkpoint update reports 200 instead of the spec's 409 here. The DO signals the no-op only through the response body, returning HTTP 200 either way. dispatch_commit is Result<()> so this arm throws the body away, and the caller returns 200 plus a cosignature.

spec says "If upload_end was too small, the mirror must respond with a '409 Conflict'." What the client currently gets is a cosignature and a 200, which means "your checkpoint is persisted and being served", while the mirror checkpoint was actually left at a larger value. So it marks the upload done and never resyncs.

Maybe you can deserialize the returned `CommittedCheckpoint` and 409 when returned.size != commit_req.size? That also gives the client a fresh t`ree_size/next_entry/ticket` to retry with.

crates/mirror_worker/src/commit.rs
```diff
@@ -0,0 +1,590 @@
...
+        TlogTile::from_index(stored_hash_index(0, last_leaf)).with_data_path(PathElem::Entries);
+    let bytes = object
+        .fetch(tile.path())
+        .await?
+        .ok_or_else(|| Error::from(format!("persisted entry bundle missing: {}", tile.path())))?;
+
+    let mut cur: &[u8] = &bytes;
+    let mut out = Vec::with_capacity(usize::try_from(count).unwrap_or(0));
+    for i in 0..count {
+        let entry = cur.read_length_prefixed(2).map_err(|e| {
```
lbaquerofierro: `read_persisted_leaves` reads count entries with no width check and no hash check, so a bundle that decodes cleanly but has wrong bytes flows into `first_package_prefix` -> `verify_package` -> subtree mismatch -> 422, which blames the client for our storage fault

crates/mirror_worker/src/add_entries.rs
lukevalenta: Good catch. Not too worried about this happening in practice, but still worth cleaning up to match the spec.

crates/mirror_worker/src/commit.rs
lukevalenta: Unlikely to happen in practice, but probably worth fixing so we don't mis-diagnose corrupted storage. (Either way, shouldn't be a correctness issue as we wouldn't persist the package.)

crates/mirror_worker/src/add_entries.rs
lukevalenta: Good catch. Fixed!

crates/mirror_worker/src/commit.rs
lukevalenta: Fixed

crates/mirror_worker/src/add_entries.rs
lukevalenta: Fixed!

crates/mirror_worker/src/commit.rs
```diff
@@ -0,0 +1,696 @@
...
+        Err(TlogError::RecordedTilesOnly) => {}
+        other => {
+            return Err(Error::from(format!(
+                "expected RecordedTilesOnly while recording hash tiles, got {other:?}"
+            )));
+        }
+    }
+
+    let mut tiles: HashMap<TlogTile, Vec<u8>> = HashMap::new();
+    for tile in recorder.0.into_inner() {
```
lbaquerofierro: Authenticating the reloaded prefix is the right call but one thing though, aren't we paying for these tiles twice? `persist_entries` turns around and calls `read_edge_tiles` at the same size and root right after, so it's mostly the same tiles again. And this loop awaits them one at a time on top of the bundle fetch, so it's a few serial round trips on every resume.

Could we `try_join_all` here at least, and maybe eventually thread one authenticated tile set through both so we only fetch once?

crates/mirror_worker/src/commit.rs
```diff
@@ -0,0 +1,696 @@
...
+    // (Re)compute and upload hash tiles.
+    for tile in TlogTile::new_tiles(persisted_size, target_size) {
+        let bytes = tile
+            .read_data(&HashReaderWithOverlay {
+                edge_tiles: &edge_tiles,
+                overlay: &overlay,
+            })
+            .map_err(|e| Error::from(format!("couldn't build hash tile {tile:?}: {e}")))?;
+        object
+            .upload(tile.path(), bytes.clone(), &immutable_tile_opts())
```
lbaquerofierro: Could we push these through `stream::iter(...).buffer_unordered(...)` instead of awaiting one at a time? Same for the bundle uploads above. Workers only gives us a handful of concurrent connections but we're using one, and since the batch here is bounded by the request body rather than a sequencing interval, a big upload also eats a lot of the subrequest budget. `apply_staged_uploads` in `log_ops` already does the `try_join_all` thing, so there's a pattern to copy. Might be worth a cap on entries per request too, with a 202 past it?

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1113 @@
...
+        tlog_witness::serialize_add_checkpoint_response(std::slice::from_ref(&note_sig));
+
+    // When the upload only reaches the mirror checkpoint's current size,
+    // the checkpoint at that size is already committed and served. There
+    // is nothing to advance, and re-committing would redundantly rewrite
+    // R2 and append a duplicate cosignature line to the served note, so
+    // return a fresh cosignature without dispatching `/commit`. (Committed
+    // is monotonic, so a stale-low snapshot only skips this optimization,
+    // never the other way.)
+    if header.upload_end <= snapshot.committed.size {
```
lbaquerofierro, line 461: Agree we shouldn't commit when nothing advances. But if we skip `/commit` here, don't we lose the only path that rewrites R2 at an unchanged size? The DO puts `COMMITTED_KEY` before writing the checkpoint and bails on failure, so if that write ever fails the served checkpoint sits behind until a *bigger* commit shows up. Before this the equal-size retry would fix it. Probably fine in practice, just want to make sure that's intentional.

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1113 @@
...
+        }
+        return Ok(snapshot.pending.clone());
+    }
+
+    // Spec: the mirror MUST also accept `upload_end` equal to the mirror
+    // checkpoint's tree size, whatever the ticket says. Everything up to
+    // it is already committed and served, so the committed checkpoint is
+    // the target and nothing new is persisted; `cosign_and_serve` returns
+    // a fresh cosignature without re-advancing.
+    if snapshot.committed.size > 0 && header.upload_end == snapshot.committed.size {
```
lbaquerofierro, line 691: This hands back `committed.signed_note_bytes`, which already has our cosignature on it. That works because the early return above skips the append, but the two are pretty far apart and nothing ties them together, so if that early return ever grows a condition we'd end up serving the note with two cosignature lines.

Would it be worth a `debug_assert!` at the append site, or keeping the bare log note around separately so the target is unambiguous?

crates/mirror_worker/src/commit.rs
lukevalenta: Ah neat, I wasn't familiar with `stream::iter::buffer_unordered`. Regarding a per-request cap on number of entries, this should be taken care of with the streaming approach in #267.

crates/mirror_worker/src/add_entries.rs
lukevalenta: Added a comment describing this behavior, but waiting for the next bigger commit should be fine. If it ends up being an issue, we could add a DO alarm to periodically check that the two are synchronized, but I don't think necessary right now.

crates/mirror_worker/src/commit.rs
lukevalenta: Addressed

crates/mirror_worker/src/commit.rs
lukevalenta: Good idea -- done!

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1125 @@
...
+    // The whole (decoded) body is buffered before processing; a follow-up
+    // commit streams it instead of buffering it all.
+    let raw = body::read_decoded_body(&parts.headers, body).await?;
+    let mut cursor = Cursor::new(raw.as_slice());
+
+    let header = match AddEntriesRequestHeader::read_from(&mut cursor) {
+        Ok(header) => header,
+        Err(e) => {
+            log::warn!("add-entries: malformed header: {e:?}");
+            return Err(AppError::BadRequest(e.to_string()));
```
lbaquerofierro: Maybe we should treat this like Eof once packages_received > 0? Right now one bad byte in a later package 400s and throws away everything we already verified, but the spec says we MUST 202 once at least one package was authenticated and saved

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1125 @@
...
+    // at all (handled above). Rejecting here would also be dishonest, since
+    // the entries were already persisted and the frontier advanced.
+
+    // When we persisted new entries, the recomputed tree MUST match the
+    // pending checkpoint the packages were proven against. A mismatch means
+    // proof verification and tile computation disagree: an internal error,
+    // never a client fault. When nothing new was persisted (a re-upload of
+    // an already-persisted range), the log-signed target is trusted
+    // directly.
+    if persisted_new && (frontier_size != header.upload_end || frontier_hash != target.hash) {
```
lbaquerofierro, line 386: I think we should run this even when `persisted_new` is false, otherwise a request that persists nothing skips it and we can end up cosigning a checkpoint at a size we never wrote tiles for.
for example,  frontier jumps 0 -> 768 from a truncated upload with committed still 0, then a client comes back with an old ticket, `upload_start = upload_end = 512` + an empty body. Nothing parses, entries is empty, this check is skipped, and /commit goes through at 512 since the DO only `checks size <= next_entry.size`.
not sure if requiring `snapshot.next_entry.hash == target.hash when upload_end == next_entry.size` would  be enough here, or if is ok for  /commit to just `require body.size == snapshot.next_entry.size`?

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1125 @@
...
+            PackageOutcome::Ok(pkg) => pkg,
+            // The body ended between or partway through a package: client
+            // truncation. Keep whatever complete packages were verified.
+            PackageOutcome::Eof => {
+                truncated = true;
+                break;
+            }
+            PackageOutcome::Err(e) => {
+                log::warn!("add-entries: malformed package [{pkg_start}, {pkg_end}): {e:?}");
+                return Err(AppError::BadRequest(e.to_string()));
```
lbaquerofierro, line 283: Maybe we should treat this like Eof once packages_received > 0? Right now one bad byte in a later package 400s and throws away everything we already verified, but the spec says we MUST 202 once at least one package was authenticated and saved

crates/mirror_worker/src/add_entries.rs
lukevalenta: Good catch! Fixed. I'm planning to add more complete integration tests and interop tests against the Sunlight implementation after this initial set of PRs is merged, but good to surface bugs like this now.

crates/mirror_worker/src/add_entries.rs
lukevalenta: Fixed!

crates/mirror_worker/src/add_entries.rs
```diff
@@ -0,0 +1,1165 @@
...
+    // Spec: the mirror updates its checkpoint to `upload_end` only once
+    // "the next entry will be greater or equal to `upload_end`", i.e. all
+    // entries up to `upload_end` are durably persisted. A request that
+    // persists nothing (e.g. an empty body, or one whose packages are all
+    // already-persisted) must not let us cosign past our frontier: without
+    // this guard `upload_end` above `next_entry` would sign a checkpoint at
+    // a size we never wrote tiles for. When the frontier has not reached
+    // `upload_end`, treat it like a truncated upload and 202 so the client
+    // resumes from the advertised next entry.
+    if frontier_size < header.upload_end {
```
lbaquerofierro, line 357: Think this fixes the opposite direction from what I meant in the last connect. This fires when the frontier is below `upload_end`, but what I was worried about was the frontier being above it.

Say `committed = 0` and `next_entry = 768`, and a client comes back with an old ticket for pending@512. Its package is all stuff we already have, so `frontier_size` stays at 768 and `persisted_new` is false. `768 < 512` is false so we skip this, skip the root check, and go cosign at 512. Problem is the 0 -> 768 commit wrote `tile/1/000.p/3` and never `.p/2`, so nobody can verify that checkpoint against our own tiles. Next bigger commit fixes the tiles but the cosignature is already out there.

Don't even need a misbehaving client, two racing normal ones can bite us here

lukevalenta: Ah, got it! That makes sense. Addressed in [65c3dde](https://github.com/cloudflare/azul/pull/264/commits/65c3dde7d9b9b7efb5ac6b0e35423ff75f30c579), which mirrors what the sunlight implementation does.

crates/mirror_worker/src/commit.rs
```diff
@@ -0,0 +1,928 @@
...
+pub(crate) async fn ensure_cut_tiles(
+    object: &impl ObjectBackend,
+    cut_size: u64,
+    frontier_size: u64,
+    frontier_hash: Hash,
+) -> Result<()> {
+    if cut_size == 0 {
+        return Ok(());
+    }
+    let base = ((cut_size - 1) / TILE_WIDTH) * TILE_WIDTH;
```
lbaquerofierro: Nice, this is better than what I was suggesting. I think there's still a gap though when the frontier is a long way ahead.

`ensure_cut_tiles` re-does just the last 256-block, so it only rewrites tiles at the levels where `base` and `cut_size` land in different tiles. For a mid tile cut that's only level 0, and the upper level partial hash tiles stay untouched, which are the ones a smaller tree wants at a different width. Say the frontier went 0 -> 2000 in one go and then we cut at 900: the only level 1 tile ever written is `.p/7`, and serving 900 needs `.p/3`. `ensure_cut_tiles_later_block` doesn't catch it because 900 and 1000 happen to sit in the same block.

A partial tile is just a prefix of a wider tile at the same index, so could we walk every level the tree at `upload_end` needs and truncate whatever's stored, instead of only redoing the last block?

lukevalenta: Ah I see. Sunlight's implementation avoids this by persisting the upper-level tiles after every entry package is parsed, but since we persist in batches we're not guaranteed that all of the partial upper-level tiles will be there. Truncating makes sense! Applied this fix.

lbaquerofierro (approved):

PR #267, opened 2026-07-30
mirror_worker: stream and incrementally commit add-entries uploads

lukevalenta:
Replace the whole-body buffer with a streaming decode: gunzip and parse the body as it arrives through a small stream-buffer adapter (retry the parse as chunks land, consume on success), so a large upload no longer holds the entire body in memory.

Persist incrementally every `commit_packages` packages (default 32, config capped at 1024) so a long or interrupted upload advances the frontier as it streams rather than only at the end. Each flush commits from the last persisted frontier and is resumable; the running frontier is threaded locally across flushes so the DO is not re-queried between chunks. Tiles are immutable and content-addressed and the DO advance is a monotone compare-and-swap, so a repeated or concurrent flush of the same range is harmless.

### DIFF
--- a/crates/mirror_worker/config.schema.json
+++ b/crates/mirror_worker/config.schema.json
@@ -37,6 +37,12 @@
             "default": 32,
             "description": "How many entry packages add-entries verifies before flushing them to storage and advancing the persisted-entry frontier. Bounds in-memory buffering and gives durable mid-request progress on large uploads. Defaults to 32 (the recommended per-request package budget) when omitted; capped at 1024 to bound worst-case buffering."
         },
+        "max_chunk_bytes": {
+            "type": "integer",
+            "minimum": 1,
+            "default": 16777216,
+            "description": "Byte ceiling on the entries buffered between flushes. add-entries flushes early once buffered entry bytes reach this many, even if fewer than commit_packages packages have accumulated. Bounds peak memory independent of package sizes, since a single package can hold up to ~16 MiB. Defaults to 16777216 (16 MiB) when omitted."
+        },
         "logs": {
             "type": "object",
             "description": "CAs this mirror mirrors, keyed by log_key_name: the CA cosigner's note-signature name (the CA ID) on the checkpoints it ingests. Used as a signed-note key name at runtime, so per c2sp.org/signed-note it MUST NOT contain '+', whitespace, or control characters.",
--- a/crates/mirror_worker/config.schema.json
+++ b/crates/mirror_worker/config.schema.json
@@ -30,6 +30,13 @@
             "default": 3600,
             "description": "How often (in seconds) the per-origin partial-tile cleaner wakes to clean orphaned partial tiles from object storage. Defaults to 3600 (one hour) when omitted."
         },
+        "commit_packages": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 1024,
+            "default": 32,
+            "description": "How many entry packages add-entries verifies before flushing them to storage and advancing the persisted-entry frontier. Bounds in-memory buffering and gives durable mid-request progress on large uploads. Defaults to 32 (the recommended per-request package budget) when omitted; capped at 1024 to bound worst-case buffering."
+        },
         "logs": {
             "type": "object",
             "description": "CAs this mirror mirrors, keyed by log_key_name: the CA cosigner's note-signature name (the CA ID) on the checkpoints it ingests. Used as a signed-note key name at runtime, so per c2sp.org/signed-note it MUST NOT contain '+', whitespace, or control characters.",
--- a/crates/mirror_worker/src/add_entries.rs
+++ b/crates/mirror_worker/src/add_entries.rs
@@ -4,19 +4,21 @@
 //! `POST /add-entries` handler.
 //!
 //! Implements the [c2sp.org/tlog-mirror `add-entries`][add-e] endpoint:
-//! read the (optionally gzip) request body, verify each [`EntryPackage`]
+//! stream the (optionally gzip) request body, verify each [`EntryPackage`]
 //! against the target pending checkpoint with a subtree consistency proof,
-//! persist the verified entries as bundles and hash tiles (see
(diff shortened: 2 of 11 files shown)


crates/mirror_worker/src/add_entries.rs
```diff
@@ -627,6 +720,106 @@ async fn advance_next_entry(env: &Env, origin: &str, size: u64, hash: Hash) -> R
...
+{
+    loop {
+        let mut cursor = Cursor::new(buf.buffered());
+        match AddEntriesRequestHeader::read_from(&mut cursor) {
+            Ok(header) => {
+                let consumed = usize::try_from(cursor.position()).unwrap_or(usize::MAX);
+                buf.consume(consumed);
+                return Ok(header);
+            }
+            Err(ParseError::Io(ref e)) if e.kind() == ErrorKind::UnexpectedEof => {
```
lbaquerofierro, line 832: Could we also retry LogOriginTruncated here? `read_from` turns an UnexpectedEof inside `log_origin` into that variant, so a valid header split partway through the origin currently gets a 400 instead of pulling the next chunk

crates/mirror_worker/src/add_entries.rs
```diff
@@ -627,6 +720,106 @@ async fn advance_next_entry(env: &Env, origin: &str, size: u64, hash: Hash) -> R
...
+) -> ApiResult<ParseOutcome>
+where
+    S: futures_util::Stream<Item = std::result::Result<Vec<u8>, BodyError>> + Unpin,
+{
+    // EOF with an empty buffer: clean truncation between packages.
+    if buf.is_eof() && buf.len() == 0 {
+        return Ok(ParseOutcome::CleanEof);
+    }
+    loop {
+        let mut cursor = Cursor::new(buf.buffered());
```
lbaquerofierro: I think it is possible to avoid restarting the package parse from byte zero after every chunk. This recopies all completed entries on each retry, so a large package arriving in small chunks can cause a lot of repeated parsing and copying before we authenticate anything

crates/mirror_worker/config.schema.json
```diff
@@ -30,6 +30,13 @@
             "default": 3600,
             "description": "How often (in seconds) the per-origin partial-tile cleaner wakes to clean orphaned partial tiles from object storage. Defaults to 3600 (one hour) when omitted."
         },
+        "commit_packages": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 1024,
+            "default": 32,
+            "description": "How many entry packages add-entries verifies before flushing them to storage and advancing the persisted-entry frontier. Bounds in-memory buffering and gives durable mid-request progress on large uploads. Defaults to 32 (the recommended per-request package budget) when omitted; capped at 1024 to bound worst-case buffering."
```
lbaquerofierro, line 38: Could we also limit this by bytes? A single package can be around 16 MiB, so buffering 32 packages could exceed the Worker's memory limit

crates/mirror_worker/src/body.rs
```diff
@@ -6,77 +6,208 @@
...
-            "Unsupported Content-Encoding: {other}"
-        ))),
     }
+
+    /// Feed one compressed chunk and return the plaintext bytes produced.
+    /// May return an empty `Vec` if `input` only completed part of the
+    /// gzip header or a DEFLATE block that hasn't emitted output yet.
+    fn push(&mut self, input: &[u8]) -> Result<Vec<u8>> {
+        self.decoder
+            .write_all(input)
```
lbaquerofierro: Is it possible to decompress this into smaller chunks? A compressed chunk can produce a much larger Vec, which is then copied again, so we could still exceed the Worker's memory limit

crates/mirror_worker/src/add_entries.rs
lukevalenta: Done!

crates/mirror_worker/src/add_entries.rs
lukevalenta: The LogOriginTruncated error doesn't actually add much over the more generic UnexpectedEof, so I've removed it.

crates/mirror_worker/config.schema.json
lukevalenta: Added a byte-based limit alongside the package-based limit.

crates/mirror_worker/src/body.rs
lukevalenta: Done!

crates/mirror_worker/src/body.rs
```diff
@@ -6,77 +6,246 @@
...
+    /// only advanced the gzip header or a not-yet-emitting DEFLATE block),
+    /// once, or many times for a highly compressible chunk.
+    ///
+    /// Input is written to the decoder one [`INFLATE_STEP`] slice at a
+    /// time, draining the sink after each, and every drained run is further
+    /// split into at most [`INFLATE_STEP`]-byte emissions. So no single
+    /// emitted run (what flows downstream) exceeds one step regardless of
+    /// the (attacker-controlled) compression ratio, and the sink is drained
+    /// per input step instead of accumulating the whole chunk's output.
+    fn push(&mut self, input: &[u8], mut emit: impl FnMut(Vec<u8>)) -> Result<()> {
```
lbaquerofierro: `INFLATE_STEP` caps the compressed bytes going in, not the plaintext coming out. `write_all` inflates the whole slice into the decoder's inner `Vec` before it returns, and deflate goes to about 1000:1. I gzipped 200 MiB of zeros and one 64 KiB slice gave me 64.3 MiB.

Same for `runs`, `push` fills it for the entire chunk before returning, so nothing flows downstream until all of it is inflated.

Maybe a separate, much smaller input step? And stop feeding once there's something to emit, parking the rest in `DecodeState`

lukevalenta: Makes sense. Addressed in [3b512ce](https://github.com/cloudflare/azul/pull/267/commits/3b512ce343d5ee7e35786d432fa65dbd7f640c54)!

lbaquerofierro (approved):

PR #271, opened 2026-07-31
tlog_core: fix rustfmt drift in cache-layer test

lukevalenta:
Reformat two assert_eq! calls that landed unformatted so `cargo fmt --all --check` is clean again. Formatting only.


lukevalenta:
Format-only change, merging.

PR #273, opened 2026-07-31
mirror_worker: add the optional sign-subtree endpoint

lukevalenta:
Implement the tlog-witness sign-subtree cosigning path: verify the requested subtree is consistent with a reference checkpoint this mirror has cosigned, then countersign the subtree with SubtreeV1CheckpointSigner.

### DIFF
--- a/crates/mirror_worker/src/frontend_worker.rs
+++ b/crates/mirror_worker/src/frontend_worker.rs
@@ -460,8 +460,8 @@ async fn sign_subtree(State(env): State<Env>, body: Bytes) -> ApiResult<axum::re
         ));
     }
 
-    // Sign the subtree. Per the spec the timestamp on a subtree
-    // cosignature MUST be zero; we use zero uniformly.
+    // Timestamp must be zero for a non-zero-start subtree; allowed for
+    // start = 0 but we use zero uniformly.
     let note_sig = subtree_signer.sign_subtree(0, origin, &subtree, &subtree_hash);
     Ok((
         StatusCode::OK,
--- a/crates/mirror_worker/src/frontend_worker.rs
+++ b/crates/mirror_worker/src/frontend_worker.rs
@@ -11,6 +11,11 @@
 //!   with one spec-mandated exception: the mirror MUST NOT cosign in
 //!   this process. Successful responses have an empty body and HTTP
 //!   status 200.
+//! - `POST /sign-subtree`: [c2sp.org/tlog-witness#sign-subtree][signsub].
+//!   Countersigns a subtree of a checkpoint this mirror has previously
+//!   cosigned. OPTIONAL in the spec but always available here, since the
+//!   mirror's cosigner is ML-DSA-44 / `subtree/v1`. Success returns the
+//!   `subtree/v1` cosignature line(s) as `text/plain`.
 //! - `GET /metadata`: mirror identity, ML-DSA-44 SPKI,
 //!   `mirror_algorithm`, prefixes, and the per-log configuration.
 //! - `GET /`: root status string.
@@ -19,6 +24,7 @@
 //! Durable Object; see [`crate::mirror_state_do`] for details.
 //!
 //! [add-cp]: https://c2sp.org/tlog-mirror#add-checkpoint
+//! [signsub]: https://c2sp.org/tlog-witness#sign-subtree
 //! [`MirrorState`]: crate::mirror_state_do
 
 use crate::{
--- a/crates/mirror_worker/src/frontend_worker.rs
+++ b/crates/mirror_worker/src/frontend_worker.rs
@@ -35,9 +35,13 @@ use axum::{
 };
 use serde::Serialize;
 use serde_with::{base64::Base64 as Base64As, serde_as};
-use signed_note::NoteError;
-use tlog_checkpoint::CheckpointText;
-use tlog_witness::{AddCheckpointRequest, CONTENT_TYPE_TLOG_SIZE, parse_add_checkpoint_request};
+use signed_note::{NoteError, NoteVerifier, VerifierList};
+use tlog_checkpoint::{CheckpointSigner as _, CheckpointText};
+use tlog_core::{Subtree, verify_subtree_consistency_proof};
+use tlog_witness::{
+    AddCheckpointRequest, CONTENT_TYPE_TLOG_SIZE, SignSubtreeRequest, parse_add_checkpoint_request,
+    parse_sign_subtree_request, serialize_sign_subtree_response,
+};
 use tower_service::Service as _;
 #[allow(clippy::wildcard_imports)]
 use worker::*;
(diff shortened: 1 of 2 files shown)


crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,10 +334,132 @@ async fn add_checkpoint(
...
+/// so this endpoint is always available.
+///
+/// Verification of the reference checkpoint is stateless: the submitted
+/// checkpoint MUST carry one of the mirror's own past `subtree/v1`
+/// cosignatures (the whole-tree cosignature it emits on a successful
+/// `add-entries`). This is safe for the same reason as in the witness: the
+/// mirror only cosigns a checkpoint after fully ingesting and verifying
+/// every entry up to that size, so a checkpoint bearing the mirror's
+/// cosignature proves the mirror holds that tree. `/sign-subtree` therefore
+/// inherits the trust window of `/add-entries`.
```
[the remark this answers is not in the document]
lukevalenta: The add-entries endpoint will be added by #264, so fine to leave as-is.

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,10 +340,132 @@ async fn add_checkpoint(
...
+    // Parse the reference checkpoint and bound-check the subtree.
+    // `Subtree::new` enforces `start < end` and the power-of-two
+    // alignment; `subtree_end <= size` is checked explicitly. Empty
+    // subtrees (`start == end`) are rejected for now; MTC draft-06 will
+    // permit them, at which point the mirror should cosign them too.
+    let cp_text = match CheckpointText::from_bytes(checkpoint.text()) {
+        Ok(t) => t,
+        Err(e) => {
+            log::warn!("sign-subtree: malformed checkpoint text: {e:?}");
+            return Err(AppError::BadRequest(format!("{e:?}")));
```
lbaquerofierro: `add_checkpoint` does this exact same parse at line 294 and uses `e.to_string()`, so it returns Bad request: malformed checkpoint text while this one returns `Bad request: MalformedCheckpointTextError. MalformedCheckpointTextError` has a `Display` impl, so could we use `e.to_string()` here too and keep the Rust type name out of the response body?

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,10 +340,132 @@ async fn add_checkpoint(
...
 // Helpers
 // ---------------------------------------------------------------------------
 
+/// Maximum `sign-subtree` request body, enforced by the route's
+/// [`DefaultBodyLimit`] layer. A well-formed request is a `subtree` range
+/// line, a base64 hash line, up to 8 subtree cosignature lines (~3.3 KiB
+/// each for ML-DSA-44), up to 63 base64 consistency-proof hash lines, and
+/// a reference checkpoint of up to `signed_note::MAX_NOTE_SIZE` (1 MiB);
+/// 1 MiB plus 64 KiB of headroom covers that.
+const MAX_SIGN_SUBTREE_BODY_SIZE: usize = 1_024 * 1_024 + 64 * 1_024;
```
lbaquerofierro: tlog_witness already exports this exact number as `MAX_REQUEST_BODY_SIZE`, and `parse_sign_subtree_request` enforces it too, so we've got the same value written down in two places that have to agree. We can just use `tlog_witness::MAX_REQUEST_BODY_SIZE`

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,10 +340,132 @@ async fn add_checkpoint(
...
+    if verify_subtree_consistency_proof(
+        &consistency_proof,
+        cp_text.size(),
+        *cp_text.hash(),
+        &subtree,
+        subtree_hash,
+    )
+    .is_err()
+    {
+        return Err(AppError::UnprocessableEntity(
```
lbaquerofierro, line 458: We're already logging the parse failure and the 403. Should we log this 422 too, plus the unknown-origin 404 and the two subtree-bounds 400s? Even just the subtree range and the checkpoint size would help, since the response body doesn't say which check failed.

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,10 +340,132 @@ async fn add_checkpoint(
...
+    // origin/size/hash with start = 0, end = size and rejects anything
+    // else.
+    let mirror_verifier: Box<dyn NoteVerifier> = subtree_signer.verifier();
+    if let Err(e) = checkpoint.verify(&VerifierList::new(vec![mirror_verifier])) {
+        match e {
+            NoteError::UnverifiedNote | NoteError::InvalidSignature { .. } => {
+                log::info!("sign-subtree: reference checkpoint not cosigned by this mirror: {e:?}");
+                return Err(AppError::ReferenceCheckpointNotCosignedByThisMirror);
+            }
+            _ => {
```
lbaquerofierro, line 435: With a one element `VerifierList` I think the only things verify can hand back are `UnverifiedNote` and `InvalidSignature`, and both are caught above. The two that would land here (`MismatchedVerifier` and `AmbiguousKey`) only happen if we built the verifier list wrong, so 400 ends up blaming the client for our bug. Not reachable today, but would `InternalServerError` be more honest?

lukevalenta: Addressed

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Good catch, fixed.

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Added the logging

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,6 +340,137 @@ async fn add_checkpoint(
...
+#[allow(clippy::too_many_lines)]
+#[worker::send]
+async fn sign_subtree(State(env): State<Env>, body: Bytes) -> ApiResult<axum::response::Response> {
+    let subtree_signer = load_mirror_signer(&env)?.as_subtree_signer();
+
+    let SignSubtreeRequest {
+        subtree_start,
+        subtree_end,
+        subtree_hash,
+        subtree_cosignatures: _,
```
lbaquerofierro, line 369: Maybe worth using `subtree_cosignatures` here as a cheap pre-screen, since the mirror checkpoint is public and otherwise anyone can reach the ML-DSA signing work? Or is rate limiting expected to handle that?

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -314,6 +340,137 @@ async fn add_checkpoint(
...
+             against checkpoint size {}",
+            cp_text.size()
+        );
+        return Err(AppError::UnprocessableEntity(
+            "subtree consistency proof failed".to_owned(),
+        ));
+    }
+
+    // Sign the subtree. Per the spec the timestamp on a subtree
+    // cosignature MUST be zero; we use zero uniformly.
```
lbaquerofierro: timestamp only has to be zero when `subtree_start` is not 0. For `subtree_start = 0`, a non-zero timestamp is allowed (just a comment updated)

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: I'm not too worried about this at this stage, but we can always add that later if we see abuse.

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Fixed!

lbaquerofierro (approved):

PR #279, opened 2026-08-26
mirror_worker: add the per-origin MirrorCleaner durable object

lukevalenta:
Adds a `MirrorCleaner` durable object, one instance per log origin, that
deletes the partial tiles the mirror leaves behind. Partials are written at
width-suffixed paths that the full tile never overwrites, so once the full
tile exists they linger as garbage and accumulate without bound. This is the
mirror's analogue of `generic_log_worker`'s `GenericCleaner`, reimplemented
for the prefixed, multi-origin bucket layout.

The full tile existing is the only thing that authorizes deleting a tile's
partials, so `clean_range` `head`-checks the unsuffixed full-tile key before
deleting anything under its `.p/` prefixes.

`add-entries` kicks the cleaner after it advances the persisted frontier, so
cleanup runs on an alarm rather than the submission path. Each wake is
bounded by the Workers subrequest limit and checkpoints its progress, so it
resumes where the previous one stopped. An origin that leaves the configured
set, most likely an MTC CA log-number window advancing, clears its own alarm
on the next wake rather than waking forever with nothing to clean.

`MirrorCleaner` is declared alongside `MirrorState` in the existing `v1`
migration rather than a new one, since `mirror_worker` has no environment
beyond `dev` and is not yet deployed.

## Cleaning ceiling

The cleaner stops at the mirror checkpoint size (`committed.size`), not the
persisted-entry frontier. `ensure_cut_tiles` writes narrow partials at
`committed.size` so the served checkpoint verifies against the mirror's own
tiles, and those partials look like orphans to the full-tile test as soon as
the frontier passes their block. Cleaning to the frontier would delete the
tiles backing the checkpoint currently being served.

The consequence is that cleaning lags cosigning rather than ingestion.

## Review notes

- Test coverage here is limited to the pure planning function `plan_clean`;
  nothing exercises the alarm loop, `clean_range`, or the subrequest budget
  end to end. The integration-test branch is next in the stack and is where
  that coverage belongs.
- Nothing guards the choice of `committed.size` over `next_entry.size`.
  Reverting that field would still pass the suite, since reading it requires
  a DO fetch and there is no mocking scaffold in this module.

### DIFF
--- a/crates/mirror_worker/src/cleaner_do.rs
+++ b/crates/mirror_worker/src/cleaner_do.rs
@@ -1,0 +1,475 @@
+// Copyright (c) 2025-2026 Cloudflare, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! `MirrorCleaner` Durable Object: deletes orphaned partial tiles from
+//! the mirror's object storage.
+//!
+//! The mirror's analogue of [`generic_log_worker`]'s `GenericCleaner`,
+//! reimplemented for the prefixed, multi-origin bucket layout (see
+//! [`crate::storage`]).
+//!
+//! One instance per configured origin (keyed by origin string, like
+//! `MirrorState`). A self-perpetuating alarm, first armed when the
+//! `add-entries` handler [kicks](kick) it after a commit, wakes every
+//! `clean_interval_secs` and, for every full tile below the mirror
+//! checkpoint size not yet cleaned, batch-deletes that tile's `.p/`
+//! partials (plus the entry-bundle partials for a level-0 tile). A
+//! per-tile `head` check confirms the full tile exists first, so a
+//! partial is never removed before its replacement lands. Progress
+//! (`cleaned_size`) is checkpointed after each deletion and each wake is
+//! bounded by the Workers subrequest limit, resuming where the last left
+//! off.
+
+use std::{cell::RefCell, mem, time::Duration};
+
+use futures_util::future::try_join_all;
+use tlog_tiles::{PathElem, TlogTile};
+#[allow(clippy::wildcard_imports)]
+use worker::*;
+
+use crate::{
+    CONFIG, MIRROR_CLEANER_BINDING,
+    mirror_state_do::{MirrorStateSnapshot, state_stub},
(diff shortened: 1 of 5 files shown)


crates/mirror_worker/src/cleaner_do.rs
```diff
@@ -0,0 +1,475 @@
...
+        }
+        for key in try_join_all(prefixes.iter().map(|prefix| self.list_prefix(prefix)))
+            .await?
+            .into_iter()
+            .flatten()
+        {
+            batch.push(key);
+            if batch.len() == MAX_DELETE_BATCH {
+                self.checked_add_subrequests(1)?;
+                self.bucket.delete_multiple(mem::take(batch)).await?;
```
lbaquerofierro: Could we keep the checkpoint behind this batch if the delete fails? `mem::take` leaves batch empty, so `clean()` can checkpoint past undeleted keys from earlier ranges

crates/mirror_worker/src/cleaner_do.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    /// List every object key under `prefix` (one page; a single tile's
+    /// partials number at most 255, well under the 1000-key page limit).
+    async fn list_prefix(&self, prefix: &str) -> Result<Vec<String>> {
+        self.checked_add_subrequests(1)?;
+        Ok(self
+            .bucket
+            .list()
+            .prefix(prefix.to_owned())
+            .execute()
```
lbaquerofierro: maybe this should keep listing while the response is truncated. R2 may return a short page, and advancing `cleaned_size` can leave partials behind

crates/mirror_worker/src/cleaner_do.rs
lukevalenta: Good catch. Fixed! (Also in the generic log cleaner)

crates/mirror_worker/src/cleaner_do.rs
lukevalenta: Fixed and also applied to the generic log cleaner.

lbaquerofierro (approved):

lbaquerofierro (approved):

PR #281, opened 2026-08-29
x509_util: accept Ed25519 chain signatures (RFC 8410)

jamestexas:
## The gap

`x509_util::is_link_valid` dispatches on the child certificate's signature algorithm OID and handles ECDSA P-256/384/521 and RSA PKCS#1 v1.5. `id-Ed25519` (1.3.101.112, RFC 8410) falls through to `_ => false`.

The effect is that an Ed25519-signed chain is rejected *as if the signature were invalid*, rather than as an unsupported algorithm — so a Static CT log built on `ct_worker` refuses every such submission at `add-chain`, and the failure doesn't say why.

Ed25519 is a standard X.509 signature algorithm and is permitted for roots and subordinate CAs, so a log whose accepted roots include an Ed25519 CA currently cannot log anything that CA issues. I hit this standing up a private Static CT log for an Ed25519-based internal CA.

## The change

One `verify_ed25519` arm alongside the existing ones, using `ed25519-dalek` — already a workspace dependency for cosignatures, so no new supply-chain surface.

It uses **`verify_strict`** rather than `verify`. That rejects non-canonical encodings and small-order public keys; the stricter reading matters more than usual for a transparency log, where signature malleability could otherwise produce two distinct log entries for one logical certificate.

## Test fixtures

An openssl-generated root (`pathlen:1`) → intermediate (`pathlen:0`) → leaf chain, all Ed25519. I confirmed the chain with

```
openssl verify -CAfile ed25519-root-ca-cert.pem \
  -untrusted ed25519-intermediate-cert.pem ed25519-leaf-cert.pem
```

before committing, so the fixtures are validated by something other than the code under test.

The new `test_validate_chain!(ed25519_chain_success; ...)` case follows the existing `p521_ecdsa_chain_success` / `sha1_rsa_chain_success` pattern. Removing the match arm fails `ed25519_chain_success` and no other test.

## Checks

- `cargo test -p x509_util` → 29 passed
- `cargo clippy -p x509_util --all-targets -- -Dwarnings -Dclippy::pedantic` → clean
- `cargo fmt --all --check` → clean

Happy to add an `add-chain` integration test against an Ed25519 root if you'd prefer the coverage at that layer too.


lukevalenta:
We originally left Ed25519 out because it's not relevant for the WebPKI (the primary use case for this project), but this is a reasonable and well-scoped change, so happy to merge.

Our focus for maintaining this project is primarily for Cloudflare's own use case, but do please let us know if you run into any other sharp corners that make it difficult to use this tool for private CA uses cases.

lukevalenta (approved):

PR #284, opened 2026-08-31
frontend_workers: remove monitoring object proxy

lukevalenta:
## Summary

- Remove catch-all R2 object routes from the CT and Bootstrap MTC submission Workers. This keeps submission Workers limited to their documented APIs and prevents their bindings from serving monitoring data directly.
- Advertise `monitoring_url` only when it is configured. Empty values are omitted instead of falling back to `submission_url`, so metadata does not claim that the submission Worker is also a monitoring endpoint. The generated configuration schemas now describe this behavior.
- Preserve integration testing against both local and deployed environments. Loopback tests read checkpoints and tiles from persisted local R2 through Wrangler, while remote tests use the advertised monitoring URL. This avoids restoring the removed frontend proxy solely for tests.
- Share the Wrangler local-R2 reader across CT, Bootstrap MTC, and mirror integration tests, and pin Wrangler 4.80.0 in CI for reproducible CLI behavior.
- Assert that persisted checkpoints return 404 through both frontend Workers, covering the removed route behavior.
- Sample checkpoint verification time after each fetch. A checkpoint can be created while it is being fetched; sampling beforehand made a valid new checkpoint appear future-dated, and reusing that timestamp caused every retry to fail.

## Verification

- `cargo test`
- `cargo fmt --all --check`
- `cargo shear`
- `cargo test -p integration_tests --test static_ct_api`
- `cargo test -p integration_tests --test bootstrap_mtc_api`
- GitHub Actions: build, lint, unit tests, semgrep, CT, Bootstrap MTC, mirror, and witness integration jobs pass

### DIFF
--- a/crates/integration_tests/tests/static_ct_api.rs
+++ b/crates/integration_tests/tests/static_ct_api.rs
@@ -10,8 +10,10 @@
 //! # Running
 //!
 //! ```text
+//! npm install --global wrangler@4.80.0
+//!
 //! # From crates/ct_worker/:
-//! npx wrangler -e=dev dev &
+//! wrangler -e=dev dev --persist-to .wrangler/state &
 //!
 //! # From workspace root:
 //! cargo test -p integration_tests --test static_ct_api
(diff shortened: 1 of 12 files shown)


crates/integration_tests/tests/static_ct_api.rs
```diff
@@ -168,6 +151,7 @@ async fn log_v3_json_returns_valid_metadata() {
         !meta.submission_url.is_empty(),
         "submission_url must be set"
     );
+    assert!(meta.monitoring_url.is_none());
```
lbaquerofierro: Deployed shards advertise `monitoring_url`, so this breaks remote runs. Can we scope this assertion to loopback URLs?

lukevalenta: Scoped this assertion to loopback runs and applied the same fix to the Bootstrap MTC metadata test. Remote runs now accept configured monitoring URLs.

mendess (approved):

PR #288, opened 2026-09-03
mirror_worker: add configurable witness service

lukevalenta:
## Summary
- add witness, mirror, and combined operating modes to `mirror_worker`
- keep witness and mirror identities, keys, and monitoring prefixes distinct
- share per-origin checkpoint state while gating role-specific routes and secrets
- remove the standalone `witness_worker`

## Stack
- depends on #287

## Testing
- `cargo test`
- workspace pedantic clippy with warnings denied
- witness and mirror standalone checks
- combined Worker release build
- live witness and mirror integration suites


crates/mirror_worker/config/src/lib.rs
```diff
@@ -1,97 +1,92 @@
...
+            Self::Mirror => "mirror",
+            Self::WitnessAndMirror => "witness-and-mirror",
+        }
+    }
+}
+
 #[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
 pub struct AppConfig {
+    pub mode: Mode,
```
mendess, line 46: can't the mode be a derived field based on the existence of `witness` and/or `mirror`? Would save some validation logic

crates/mirror_worker/config/src/lib.rs
```diff
@@ -1,97 +1,92 @@
...
+    pub witness: Option<IdentityConfig>,
+    pub mirror: Option<MirrorConfig>,
+    #[serde(deserialize_with = "deserialize_logs")]
+    pub logs: HashMap<String, LogParams>,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct IdentityConfig {
+    pub name: String,
```
mendess, line 59: Could this be `: KeyName` instead and roll up the validation into the deserialization? This would also reduce the amount of validation logic in `AppConfig::validate`

crates/mirror_worker/config/src/lib.rs
```diff
@@ -1,97 +1,92 @@
...
-    /// [tiles]: https://c2sp.org/tlog-tiles
     pub monitoring_prefix: Option<String>,
-    /// How often (in seconds) the per-origin partial-tile cleaner wakes
-    /// to clean orphaned partial tiles from object storage. `None` falls
-    /// back to a one-hour default (see [`Self::clean_interval_secs`]).
-    /// Consumed by [`mirror_worker`](../mirror_worker/)'s `cleaner_do`.
+    pub witness: Option<IdentityConfig>,
+    pub mirror: Option<MirrorConfig>,
+    #[serde(deserialize_with = "deserialize_logs")]
+    pub logs: HashMap<String, LogParams>,
```
mendess, line 53: the key could also be of type `KeyName`, right?

crates/mirror_worker/src/add_entries.rs
```diff
@@ -1014,27 +1007,36 @@ fn resolve_target_pending(
     snapshot: &MirrorStateSnapshot,
     verifiers: &signed_note::VerifierList,
 ) -> std::result::Result<PendingCheckpoint, &'static str> {
-    if header.upload_end == snapshot.pending.size {
+    let Some(pending) = snapshot.pending.as_ref() else {
```
mendess, line 1010: nit: can be shorter

```suggestion
    let Some(pending) = &snapshot.pending else {
```

crates/mirror_worker/src/frontend_worker.rs
```diff
@@ -158,404 +133,319 @@ impl From<worker::Error> for AppError {
...
+        AppError::BadRequest(error.to_string())
+    })?;
+    let (old_size, consistency_proof, checkpoint, checkpoint_text) = validated.into_parts();
+    let origin = checkpoint_text.origin();
     let Some(verifiers) = log_verifiers(origin) else {
         return Err(AppError::UnknownLogOrigin);
     };
+    verify_source_checkpoint(&checkpoint, &verifiers, "add-checkpoint")?;
+
+    let witness_signature = if enabled_roles(CONFIG.mode).witness() {
```
lbaquerofierro, line 315: Could we generate the witness signature only after `dispatch_update_pending` succeeds? A client can replay any public, valid checkpoint with a stale `old_size` and make us perform an ML-DSA signature before returning 409

crates/mirror_worker/config/src/lib.rs
```diff
@@ -1,97 +1,92 @@
...
-    /// served at `<monitoring_prefix>/<encoded origin>/...`). `None`
-    /// means "same as `submission_prefix`".
-    ///
-    /// [tiles]: https://c2sp.org/tlog-tiles
     pub monitoring_prefix: Option<String>,
-    /// How often (in seconds) the per-origin partial-tile cleaner wakes
-    /// to clean orphaned partial tiles from object storage. `None` falls
-    /// back to a one-hour default (see [`Self::clean_interval_secs`]).
-    /// Consumed by [`mirror_worker`](../mirror_worker/)'s `cleaner_do`.
+    pub witness: Option<IdentityConfig>,
```
lbaquerofierro, line 50: Combined mode needs separate monitoring prefixes for the witness and mirror. The tlog-mirror spec requires them to be distinct, and the checkpoint written by add-entries only includes the mirror signature, so this prefix cannot also serve the witness checkpoint. Could we have a prefix per role and keep the witness-signed checkpoint separately?

crates/mirror_worker/src/mirror_state_do.rs
```diff
@@ -230,51 +216,47 @@ impl MirrorState {
...
+                    body.old_size,
+                    CheckpointState {
+                        size: body.new_size,
+                        hash: body.new_hash,
+                    },
+                    &body.proof,
+                );
+                if let Err(error) = transition {
+                    return match error {
+                        CheckpointTransitionError::OldSizeMismatch
```
lbaquerofierro, line 235: Consistency failures should return 422, righ tnow same size hash mismatches return 409, while non-empty initial or same-size proofs and invalid empty-tree hashes return 400. Could we reserve 409 for OldSizeMismatch and map these other transition errors to 422?

crates/mirror_worker/config/src/lib.rs
lukevalenta: Oh yeah nice simplification.

crates/mirror_worker/config/src/lib.rs
lukevalenta: Addressed in fc533a7

crates/mirror_worker/config/src/lib.rs
lukevalenta: Addressed in 9442eb7

crates/mirror_worker/config/src/lib.rs
lukevalenta: Addressed in fc533a7

crates/mirror_worker/src/add_entries.rs
lukevalenta: Addressed in 8c21808

crates/mirror_worker/config/src/lib.rs
lukevalenta: Addressed in 75f08d1

crates/mirror_worker/src/frontend_worker.rs
lukevalenta: Addressed in 49edb38

crates/mirror_worker/src/mirror_state_do.rs
lukevalenta: Addressed in 468ccc9

crates/mirror_worker/config/src/lib.rs
lbaquerofierro: This separates the advertised prefixes, but it still does not persist a checkpoint for the witness prefix. `add-checkpoint` stores the original log-signed note, and `add-entries` later stores that note with only the mirror cosignature, so the witness monitoring endpoint has no witness-cosigned checkpoint to serve

crates/mirror_worker/config/src/lib.rs
```diff
@@ -126,541 +103,317 @@ where
...
-    /// [`deserialize_logs`]). A plain `serde_json` object silently keeps
-    /// only the last value for a repeated key, so without that check a
-    /// duplicate `log_key_name` would drop a trusted CA (its keys and
-    /// log-number window) with no error.
+    /// Panics if mirror configuration is absent.
+    pub fn mirror_config(&self) -> &MirrorConfig {
+        self.mirror
+            .as_ref()
+            .expect("validated mirror mode must have mirror config")
+    }
```
mendess: why not return an Option<&MirrorConfig> here and remove the `mirror_enabled` function? `Option::as_ref` and `Option::is_some` are `const fn` so `AppConfig::mirror_enabled` can be achieved with `config.mirror_config().is_some()` if needed in const contexts and this way you can remove an `.expect` call

mendess (changes requested):

crates/mirror_worker/config/src/lib.rs
lukevalenta: Addressed in e9bb8f7

crates/mirror_worker/config/src/lib.rs
lukevalenta: Oops, fixed in 286e076! This just adds a separate R2 bucket for the witness to write its checkpoints.

mendess (approved):

crates/mirror_worker/src/mirror_state_do.rs
```diff
@@ -224,94 +237,145 @@ impl MirrorState {
...
+        if CONFIG.witness_enabled()
+            && let Some(checkpoint) = current.as_mut()
+            && !checkpoint.witness_published
+        {
+            self.recover_witness_checkpoint(checkpoint).await?;
+        }
+        if CONFIG.witness_enabled()
+            && let Some(checkpoint) = current.as_ref()
+            && !checkpoint.witness_response_bytes.is_empty()
+            && (checkpoint.update_request_hash == request_hash
```
lbaquerofierro, line 265: This legacy state fallback runs before transition validation,a stale old size or invalid proof can return 200 instead of 409/422. Can we validate the transition before applying this fallback and add a migration test?

lukevalenta: Addressed in https://github.com/cloudflare/azul/pull/294

PR #289, opened 2026-09-03
mirror_worker: recover interrupted checkpoint publication

lukevalenta:
## Summary
- persist checkpoint publication intent before writing the R2 object
- complete interrupted publication before processing subsequent commits
- make same-size retries repair absent checkpoint objects

## Stack
- depends on #288

## Testing
- `cargo test`
- `cargo fmt --all --check`
- `cargo shear`
- live mirror integration suite


crates/mirror_worker/src/add_entries.rs
```diff
@@ -593,25 +593,14 @@ async fn cosign_and_serve(
     // The served object includes the mirror cosignature returned below.
     let checkpoint_obj = [target.signed_note_bytes.as_slice(), &cosig_body].concat();
 
-    if snapshot
```
lukevalenta, line 596: TODO: double-check this carefully, since we iterated a bit on this behavior in a previous PR

lukevalenta: This just defers the check to the DO so it supports retrying same-size commits.

crates/mirror_worker/src/mirror_state_do.rs
```diff
@@ -367,6 +407,23 @@ impl MirrorState {
...
+) -> Result<()>
+where
+    O: ObjectBackend,
+    F: Future<Output = Result<()>>,
+    G: Future<Output = Result<()>>,
+{
+    commit::write_checkpoint(object, checkpoint.signed_note_bytes.clone()).await?;
+    persist.await?;
+    clear.await
+}
```
mendess, line 425: this function is a bit odd, it takes any future. Could it instead be something like:

```rs
async fn finish_publication<O, F, G>(
    object: &O,
    checkpoint: &CommittedCheckpoint,
    storage: &Storage,
) -> Result<()>
where
    O: ObjectBackend,
    F: Future<Output = Result<()>>,
    G: Future<Output = Result<()>>,
{
    commit::write_checkpoint(object, checkpoint.signed_note_bytes.clone()).await?;
    storage.put(COMMITTED_KEY, &object).await?;
    storage.delete(PUBLISHING_KEY).await?;
    Ok(())
}
```

and called like
```rs
        finish_publication(&bucket, &new_committed, &storage).await?;
```

lukevalenta: Addressed in 805765c

lukevalenta: Addressed in 09e130b

mendess (approved):

